PR #4, opened 2016-07-07
Use normalize_whitespace filter for meta description

pathawks:
Taking advantage of jekyll/jekyll#4917 to solve jekyll/jekyll#5066

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -4,7 +4,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
 
   <title>{% if page.title %}{{ page.title | escape }}{% else %}{{ site.title | escape }}{% endif %}</title>
-  <meta name="description" content="{% if page.excerpt %}{{ page.excerpt | strip_html | strip_newlines | truncate: 160 }}{% else %}{{ site.description | escape }}{% endif %}">
+  <meta name="description" content="{{ page.excerpt | default: site.description | strip_html | normalize_whitespace | truncate: 160 | escape }}">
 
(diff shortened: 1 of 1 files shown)


_includes/head.html
```diff
@@ -4,7 +4,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
 
   <title>{% if page.title %}{{ page.title | escape }}{% else %}{{ site.title | escape }}{% endif %}</title>
-  <meta name="description" content="{% if page.excerpt %}{{ page.excerpt | strip_html | strip_newlines | truncate: 160 }}{% else %}{{ site.description | escape }}{% endif %}">
+  <meta name="description" content="{% if page.excerpt %}{{ page.excerpt | strip_html | normalize_whitespace | truncate: 160 }}{% else %}{{ site.description | escape | normalize_whitespace }}{% endif %}">
```
pathawks: Is there some reason for the two different filter chains? Why not:

```text
<meta name="description" content="{{ page.excerpt | default: site.description | strip_html | normalize_whitespace | truncate: 160 }}">
```

parkr: `site.description` is also run through `escape`. Other than that, I see no big reason they should be separate. `page.excerpt` could be run through `escape` as well.

pathawks:
Updated to get rid of the `if` :+1:

parkr:
LGTM.

parkr:
LGTM.

parkr:
@jekyllbot: merge

PR #6, opened 2016-07-28
include _layouts, _includes, and _sass dir in gem

Crunch09:
fixes jekyll/jekyll#5145.

before:
screen shot 2016-07-28 at 12 06 36

### DIFF
--- a/minima.gemspec
+++ b/minima.gemspec
@@ -12,7 +12,7 @@ Gem::Specification.new do |spec|
 
   spec.metadata["plugin_type"] = "theme"
 
-  spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(exe|_layouts|_includes|_sass)/}) }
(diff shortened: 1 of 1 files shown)


benbalter:
LGTM. We'll need to also fix this [upstream](https://github.com/jekyll/jekyll/blob/master/lib/theme_template/theme.gemspec.erb#L13).

Crunch09:
Ok, would be happy to update this there as well, if no one has already started.

benbalter:
@Crunch09 opened https://github.com/jekyll/jekyll/pull/5152, but realized that might not the right approach either, as we may want to include, e.g., the LICENSE. Was just a quick fix, so please feel free to open a PR if you have a better approach.

Crunch09:
@benbalter just saw it 👍 will try to have a more closer look later.

parkr:
`.reject.match` never caught by anyone before. Love it.

@jekyllbot: merge

PR #21, opened 2016-08-25
Add support to Disqus

nicholaspufal:
This PR enables the user to show, optionally, Disqus comments below each post.

In order to get this working the user only needs to add the following to `_config.yml`:

```yaml
disqus:
  shortname: my_disqus_shortname
```

You can find out more about Disqus' shortnames [here](https://help.disqus.com/customer/portal/articles/466208).

The current behavior remains the same if no `disqus` property is supplied.

### DIFF
--- a/_includes/disqus_comments.html
+++ b/_includes/disqus_comments.html
@@ -1,0 +1,19 @@
+{% if page.comments %}
+  <div id="disqus_thread"></div>
+  <script>
+    var disqus_config = function () {
+      this.page.url = '{{ site.url }}{{ page.url }}';        // Replace PAGE_URL with your page's canonical URL variable
+      this.page.identifier = '{{ site.url }}{{ page.url }}'; // Replace PAGE_IDENTIFIER with your page's unique identifier variable
+    };
+
+    (function() {  // DON'T EDIT BELOW THIS LINE
+      var d = document, s = d.createElement('script');
+
+      s.src = '//{{ site.disqus.shortname }}.disqus.com/embed.js';
+
+      s.setAttribute('data-timestamp', +new Date());
+      (d.head || d.body).appendChild(s);
+    })();
+  </script>
+  <noscript>Please enable JavaScript to view the <a href="https://disqus.com/?ref_noscript" rel="nofollow">comments powered by Disqus.</a></noscript>
+{% endif %}
--- a/_includes/disqus_comments.html
+++ b/_includes/disqus_comments.html
@@ -2,11 +2,11 @@
   <div id="disqus_thread"></div>
   <script>
     var disqus_config = function () {
-      this.page.url = '{{ site.url }}{{ page.url }}';        // Replace PAGE_URL with your page's canonical URL variable
-      this.page.identifier = '{{ site.url }}{{ page.url }}'; // Replace PAGE_IDENTIFIER with your page's unique identifier variable
+      this.page.url = '{{ site.url }}{{ page.url }}';
+      this.page.identifier = '{{ site.url }}{{ page.url }}';
     };
 
-    (function() {  // DON'T EDIT BELOW THIS LINE
+    (function() {
       var d = document, s = d.createElement('script');
 
       s.src = '//{{ site.disqus.shortname }}.disqus.com/embed.js';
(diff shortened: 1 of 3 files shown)


DirtyF:
In order to avoid unnecessary requests in development mode, comments could only be displayed in production mode (` {% if jekyll.environment == "production" %}`…)

_includes/disqus_comments.html
```diff
@@ -0,0 +1,19 @@
...
+  <script>
+    var disqus_config = function () {
+      this.page.url = '{{ site.url }}{{ page.url }}';
+      this.page.identifier = '{{ site.url }}{{ page.url }}';
+    };
+
+    (function() {
+      var d = document, s = d.createElement('script');
+
+      s.src = '//{{ site.disqus.shortname }}.disqus.com/embed.js';
```
DirtyF: > Protocol relative URL is considered an anti pattern, always use https:// if asset is available on SSL ([source](http://www.paulirish.com/2010/the-protocol-relative-url/))

_includes/disqus_comments.html
```diff
@@ -0,0 +1,19 @@
+{% if page.comments %}
+  <div id="disqus_thread"></div>
+  <script>
+    var disqus_config = function () {
+      this.page.url = '{{ site.url }}{{ page.url }}';
+      this.page.identifier = '{{ site.url }}{{ page.url }}';
```
DirtyF: What if `{{ site.github.url }`} is defined instead of `{{ site.url }}` ?

nicholaspufal: As part of my newest commit (https://github.com/jekyll/minima/pull/21/commits/7f5ec66c334555b0155549c1d07adaf1a09f9030) I changed this behavior. Please let me know if that's acceptable. 

I don't have much experience with Jekyll + GH Pages, but I'm assuming if `site.url` is not present then I can use `site.github.url` as a fallback (?)

nicholaspufal:
I made all the suggested changes. Let me know if it looks good or if I should update anything else :)

_includes/disqus_comments.html
```diff
@@ -0,0 +1,25 @@
+{% if page.comments and jekyll.environment == "production" %}
```
benbalter: Does this require every post to have `comments: true?` perhaps make it opt out for posts?

benbalter:
@nicholaspufal this is looking good! Mind adding some documentation to the README on how to add comment support to my site?

_includes/disqus_comments.html
nicholaspufal: :like: I just made that flag optional, meaning that comments will be showed by default

nicholaspufal:
@benbalter I just added a couple of lines to the README file explaining how to use it. Let me know if I missed anything or we are good to go 👍

benbalter:
Question: should the theme add comments to *pages* by default? E.g., about?

DirtyF:
@benbalter Good point. We should add comments only on pages with post layout by default.

benbalter:
> We could add comments only on pages with post layout by default?

We could, but that's a really complicated out-of-the-box experience for what I suspect, is the default/most common desired behavior. I guess put it another way:

1. Do we want pages to have comments by default? Do we think most users want this? if not, do we want to force the user to manually disable it?

2. Do we want posts to have comments by default? Do we think most users want this? If so, do we want to force users to manually enable it?

I suspect that the majority of users, if they enabled discuss, would want comments on posts (e.g., a blog post), but not Pages (e.g., an about page, their index page), or any other category (which I *think* is the current behavior, since the include is only included on posts).

nicholaspufal:
I agree with all points listed here. As @benbalter already mentioned, the current behavior already is this: 

IF the user enabled disqus (meaning they added a new property to their `_config.yml` file) and IF it's a post (that `_disqus_comments.html` partial is only being included as part of the `post.html` layout) then it will show comments by default.

If they haven't added disqus then no comments would be displayed as the partial won't be loaded at all.

If it's not a `post` then it will not load that partial as well, meaning no comments will be displayed.

In my mind the current behavior is already what you guys mentioned, am I correct?

DirtyF:
LGTM

benbalter:
LGTM.

benbalter:
@jekyllbot: merge +major

benbalter:
Thanks @nicholaspufal!

PR #22, opened 2016-08-28
Added "author" in footer instead of double title

Lorthirk:
I wrote this PR because I thought that having site.title shown twice in the footer was redundant, so I introduced the "author" variable in _config.yml to be shown just above the email address. Moreover, this behavior is optional, since if you don't define the "author" variable it falls back to the "title" variable.


benbalter:
LGTM.

Thanks @Lorthirk!

PR #23, opened 2016-08-30
Add logo to the header

ashmaroli:
Users should be able to add their logo easily to their blog.
With this PR all they need to do is replace the logo.png in the assets folder or simply remove the `logo:` key from the config file to revert to old behaviour.


_includes/header.html
```diff
@@ -2,7 +2,13 @@
 
   <div class="wrapper">
 
-    <a class="site-title" href="{{ site.baseurl }}/">{{ site.title | escape }}</a>
+    <a class="site-title" href="{{ site.baseurl }}/">
+      {% if site.logo %}
+        <img src="/assets/{{ site.logo }}" class="site-logo" width="150" height="auto">
```
DirtyF: Should we encourage people to put their images in `/assets/img/`?

DirtyF:
@ashmaroli Could you document this in the README and explain how to remove the logo for people who don't wish to display a logo?

_includes/header.html
ashmaroli: How does using JekyllAssets plugin work? Will it create conflicts with this change?

ashmaroli:
@DirtyF, sure, I'll push an update to documentation too.

_includes/header.html
DirtyF: AFAIK `jekyll-assets` is not embedded in Jekyll by default and will process `_assets/img/` or `assets/images` but I don't see any reason it could create a conflict. cc/ @envygeeks

ashmaroli: I've decided that this PR will stick to adding the logo to /assets over /assets/images. Once /assets get a bit crowded with other types of files, we'll segregate to individual folders.

ashmaroli:
Now that I've added the documentation, I feel this should only be merged close to the next release, to avoid confusion, in case someone refers the doc during the interim.

_includes/header.html
envygeeks: We copy directly into `/assets` in the `_site` folder which might affect the site as far as we would not preserve the images and they would always be recopied and in development if a file in `_assets` conflicts with a file in `/assets` we would win and overwrite their copy, though this would only be a problem in development or if the user explicitly disabled SHA's.

ashmaroli: :confused:  
so..... what do you recommend I do?

envygeeks: There is really nothing you can do.  I saw this situation coming when this feature was made and tried to make suggestions to ease it but those suggestions were not accepted so all you can do is keep on keeping on until it becomes a problem.

ashmaroli: > so all you can do is keep on keeping on 

???

@DirtyF, so.. hw abt I change the path to have the logo at the root instead?

ashmaroli:
Altering logo path to be at the root of the blog..

jglovier:
I'm not totally in favor of this primarily because the `jekyll new` theme is intended for personal blog use, and a logo implies a company type of blog, which would typically mean you would not use most of the this theme (i.e. most companies would want to make their own theme entirely if they have a logo and branding).

That being said, I'm not entirely opposed to it either, given your approach of adding it as a non-default option. 

The biggest concern is just making sure that when it's enabled, it doesn't break the site header styling. The site header wasn't designed to accommodate a logo in the first place, so I'm afraid adding a logo to this theme as is will feel awkward in practice.

_includes/header.html
```diff
@@ -2,7 +2,13 @@
 
   <div class="wrapper">
 
-    <a class="site-title" href="{{ site.baseurl }}/">{{ site.title | escape }}</a>
+    <a class="site-title" href="{{ site.baseurl }}/">
+      {% if site.logo %}
+        <img src="/{{ site.logo }}" class="site-logo" width="150" height="auto">
```
jglovier: For this to not break the site header layout you should replace the width attribute with a `height="56"` attribute.

ashmaroli: okay, I'll switch the attributes and set the height to match `site-title { line-height: 56px; }`

ashmaroli:
> a logo implies a company type of blog

@jglovier Agreed, but not all the time. Bloggers love having graphics in their headers (usually an entire graphic header).
 Business / Corporate blogs would have a logo and additionally a set style-guide in sync with their brand.

> The site header wasn't designed to accommodate a logo

True. But we can provide an option for those interested, or at least show them one way of getting it done. :smiley:

jglovier:
>Bloggers love having graphics in their headers (usually an entire graphic header).

True, but the aim of this particular theme is to very intentionally not be that type of blog. The intent of the design is very intentionally minimal and typography based. Trying to make it fit other styles is beyond the scope of what I designed this theme to be.

>True. But we can provide an option for those interested, or at least show them one way of getting it done.

Yeah, as long as we can address the aforementioned concerns, I think that's fair.

jglovier:
@ashmaroli could you post a screencap of the changed version including the logo as it renders on desktop and mobile widths for reference and posterity?

pathawks:
This seems to assume that the logo contains the textual site name that is legible at a (relatively) small size.

ashmaroli:
## Screencaps
### Desktop
desktop-res

desktop post

### Tablet (iPad)
ipad-scr

### Mobile (iPhone 5)
iphone-scr

ashmaroli:
@pathawks initially I made no such assumption. The header was free to expand in height.
With Joel's suggestion in place, it now assumes the logo be legible at a height of 56px. 
I'm thinking of updating the documentation to explicitly mention this:

```
NOTE: minima supports only logo images with a max-height of 56px
```

README.md
```diff
@@ -22,7 +22,18 @@ And then execute:
 
     $ bundle
 
-## Enabling comments (via Disqus)
+## Customizing
+
+### Logo
+
+By default, `jekyll new` doesn't generate sites with a logo image included. To add a logo when using minima, simply add the full filename (incl. extn.) to your *_config.yml* like:
```
benbalter: We should write out `including` and `extension`.

_includes/header.html
```diff
@@ -2,7 +2,13 @@
 
   <div class="wrapper">
 
-    <a class="site-title" href="{{ site.baseurl }}/">{{ site.title | escape }}</a>
+    <a class="site-title" href="{{ site.baseurl }}/">
+      {% if site.logo %}
+        <img src="/{{ site.logo }}" class="site-logo" width="auto" height="56">
```
benbalter: This assumes the site is at the root. We'll need to prepend `site.baseurl`.

benbalter:
Agree with much of the concern @jglovier expressed. If a publisher is sophisticated enough to have their own logo, they're generally going to be more sensitive to branding, and Minima may not be a good fit for theme. Put another way, corporate blogs aren't what Minima is intended for — they're not our target user — and I'm worried the user experience of our target user would suffer (users that need help creating a themed site), might suffer. But also as @jglovier pointed out, so long as it's disabled by default, and doesn't add too much complexity to the template, that risk should be minimized.

ashmaroli:
Hello all, is this pull going to be up for consideration? 
I'll rebase if yes, otherwise I'll close.. :smiley:

parkr:
I personally feel a logo is beyond the scope of this project. Thanks for the idea, but I don't think minima is the right place for this kind of feature. Thanks!

PR #25, opened 2016-08-30
Look for site.github.url if it exists

### LINKED ISSUE #17: Site should look to `site.github.url` if it exists, and if not, fall back to `baseurl`
This would allow project pages to work out of the box on GitHub Pages without requiring additional configuration (and wouldn't upset the local development experience or the experience of those using Jekyll outside of GitHub Pages).

parkr:
This fixes #17. /cc @benbalter

This requires a change in the `jekyll new` template, sadly, namely in the listing of `site.posts`. We could potentially ship an `include` for `posts_list` or something so we don't have to update it in Jekyll in the future?

### DIFF
--- a/example/_config.yml
+++ b/example/_config.yml
@@ -1,11 +1,17 @@
-theme: minima
 title: Your awesome title
-email: your-email@domain.com
 author: Mr. GitHub User
-description: "Write an awesome description for your new site here. It will appear in your document head meta (for Google search results) and in your feed.xml site description."
+email: your-email@domain.com
+description: > # this means to ignore newlines until "baseurl:"
+  Write an awesome description for your new site here. You can edit this
+  line in _config.yml. It will appear in your document head meta (for
+  Google search results) and in your feed.xml site description.
+baseurl: "/minima"
 twitter_username: jekyllrb
 github_username:  jekyll
 
-collections:
-  posts:
-    permalink: /posts/:title
+# Build settings
+markdown: kramdown
+theme: minima
+exclude:
+  - Gemfile
+  - Gemfile.lock
--- a/example/index.html
+++ b/example/index.html
@@ -1,10 +1,13 @@
 ---
-layout: default
+layout: page
 ---
 
 <div class="home">
 
   <h1 class="page-heading">Posts</h1>
+
+  {% assign custom_url = site.url | append: site.baseurl %}
+  {% assign full_base_url = site.github.url | default: custom_url %}
 
   <ul class="post-list">
     {% for post in site.posts %}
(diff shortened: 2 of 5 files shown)


example/index.html
```diff
@@ -1,23 +1,26 @@
...
-layout: default
+layout: page
 ---
 
 <div class="home">
 
   <h1 class="page-heading">Posts</h1>
 
+  {% assign custom_url = site.url | append: site.baseurl %}
+  {% assign full_base_url = site.github.url | default: custom_url %}
```
benbalter: If the user sets a base url, wouldn't this result in `http://username.github.io/project/project/`?

benbalter: We need to assume users are going to look to the default theme as an example of how they should build their own site... I'd really rather avoid this complexity if possible. Is there any way to do this without assigns?

parkr: > If the user sets a base url, wouldn't this result in http://username.github.io/project/project/?

What do you mean by `this`? `custom_url` is `site.url + site.baseurl`, and `full_base_url` is `site.github.url` OR `custom_url`, if `site.github.url` doesn't exist.

> Is there any way to do this without assigns?

We need to allow plugins or provide a filter like this in Jekyll or the jekyll-github-metadata plugin. Due to the scope at render time (starts with just the page content, then includes each successive layout in turn), we have to calculate this each time. Alternatively, Jekyll could introduce `site.full_url`.

benbalter:
LGTM.

ashmaroli:
ever since this got merged, I'm no longer able to access `http://localhost:4000`
I have to instead go to `http://localhost:4000/minima/index` to reach the `example` site

/cc @parkr, @benbalter

example/_config.yml
```diff
@@ -1,11 +1,17 @@
...
 title: Your awesome title
-email: your-email@domain.com
 author: Mr. GitHub User
-description: "Write an awesome description for your new site here. It will appear in your document head meta (for Google search results) and in your feed.xml site description."
+email: your-email@domain.com
+description: > # this means to ignore newlines until "baseurl:"
+  Write an awesome description for your new site here. You can edit this
+  line in _config.yml. It will appear in your document head meta (for
+  Google search results) and in your feed.xml site description.
+baseurl: "/minima"
```
benbalter: @parkr did you mean to include a baseurl by default?

parkr: No.

parkr: Well.

parkr: It is useful for testing.

ashmaroli: we'll hv update readme to reflect this change..

anthonyjsmith:
I think that having a full URL in the stylesheet link causes a default Jekyll site to look for a stylesheet at `http://example.com/css/main.css`, which obviously doesn't exist. See [this Jekyll Talk thread](https://talk.jekyllrb.com/t/jekyll-build-does-not-create-complete-default-directory-structure/2953/2).

Issue #29, opened 2016-09-05
Gemspec file doesn't include entire theme


kulinacs:
The following line:
spec.files = `git ls-files -z`.split("\x0").select { |f| f.match(%r{^(_includes|_layouts|_sass|LICENSE|README)/i}) }
causes the gem to only package the files:
_includes/icon-github.html
_includes/icon-github.svg
_includes/icon-twitter.html
_includes/icon-twitter.svg

PR #31, opened 2016-09-06
Fix gemspec regex

benbalter:
Fixes https://github.com/jekyll/minima/issues/29.

It's a quick fix to get the files added, but I'm hoping to fix this upstream and include it as a constant so that other themes can reference it.

### DIFF
--- a/minima.gemspec
+++ b/minima.gemspec
@@ -12,7 +12,10 @@ Gem::Specification.new do |spec|
 
   spec.metadata["plugin_type"] = "theme"
 
-  spec.files         = `git ls-files -z`.split("\x0").select { |f| f.match(%r{^(_includes|_layouts|_sass|LICENSE|README)/i}) }
+  spec.files         = `git ls-files -z`.split("\x0").select do |f|
+    f.match(%r{^(_(includes|layouts|sass)/|(LICENSE|README)((\.(txt|md|markdown)|$)))}i)
+  end
+  
   spec.bindir        = "exe"
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
 


ashmaroli:
There's a problem with this regex.. it doesn't include the following:
- Gemfile
- .gemspec
- .gitignore 

Since there's more to include than reject, I propose a reject expression instead:
```
  spec.files = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
```

benbalter:
@ashmaroli I believe the Gemfile and Gemspec is only used for local development (gemspec's aren't executed by downstream consumers), and since the resulting Gem is not a git repository, the gitignore file is no longer necessary.

ashmaroli:
alright then..
but I still think a reject exp is simpler than the select exp

benbalter:
> but I still think a reject exp is simpler than the select exp

Simpler for us as Jekyll maintainers, yes. But see my comment over in https://github.com/jekyll/jekyll/pull/5325#issuecomment-245065267. I'm proposing that we make it *harder* for us, in order to make it easier for theme creators (by making it easier to do the right thing and harder to do the wrong thing, by default).

ashmaroli:
> Simpler for us as Jekyll maintainers

I meant simpler for any human reading the .gemspec file. But I agree with your point that **we want the default to be a whitelist, not a blacklist.**
So, :+1:

pathawks:
Gemspecs are not meant to be human readable. That’s why good documentation is so important :+1:

benbalter:
@ashmaroli realize tone can not always come across the internet and just wanted to be clear, I really appreciate you taking the time to fix this and appreciate our extremely productive discussion even more — thank you.

benbalter:
I'm going to merge this here, so that we can release a `1.1.0` to fix issues with sites not at the domain root not being styled, but would very much like to continue the discussion in https://github.com/jekyll/jekyll/issues/5323 upstream (and @ashmaroli's related PR) to figure out the best long-term solution;

benbalter:
@jekyllbot: merge

Issue #16, opened 2016-08-14
Theme hardcodes CSS path to `css/main.css`


benbalter:
That means that if I'm adding minima to a site, I have to create an entire CSS folder, even if I only have one CSS file. 

Instead, it should be `assets/style.scss`(or similar) where the downstream users may have a use for storing other assets in that folder.

benbalter:
Is there any way to implement this without breaking existing sites?

davidyell:
I found this because I was constantly getting an error with a missing css file because it Jekyll was looking in `assets/main.css` instead of `css/main.css`. I've had to overload the theme in my local project to solve this.

If this is the default, then everyone who uses Jekyll for the first time will always get a site without any css.

ashmaroli:
This issue has been resolved, is it not?

ashawley:
Yeah, seems this was fixed in #43 and released in 2.0?

PR #43, opened 2016-09-23
move `css/` directory from jekyll into `assets/`

ashmaroli:
now that jekyll supports `assets/` dir within theme gems, `css/` and its
contents will be housed within minima's own `assets/` directory

Ref: jekyll/jekyll#5402


_includes/head.html
```diff
@@ -8,7 +8,7 @@
 
   {% assign custom_url = site.url | append: site.baseurl %}
   {% assign full_base_url = custom_url | default: site.github.url %}
-  <link rel="stylesheet" href="{{ "/css/main.css" | prepend: full_base_url }}">
+  <link rel="stylesheet" href="{{ "/assets/css/main.css" | prepend: full_base_url }}">
```
pathawks: Can we keep the permalink of the CSS  the same?

ashmaroli: Are you asking me? or are you requesting me to undo this?

ashmaroli: either ways, I think the answer is no.. its going to be either `assets/css/main.css` or `assets/main.css` if we're to follow Parker's Game Plan for Jekyll 3.3

benbalter: For context, the idea here is that `/assets` will be a magic folder that will be included in the downstream site, saving the user from needing to create a dedicated `scss` file. See https://github.com/jekyll/minima/issues/16.

For existing users, it should "just work", but will need to require Jekyll `~> 3.3` in order to ensure we're not breaking existing sites.

pathawks: Can a `page` in `assets/` not have a custom permalink in YAML?

ashmaroli: > in order to ensure we're not breaking existing sites.

from my understanding, this is a breaking change. So should an existing site use minima 2.0, they have to make this change as well..

benbalter: Huh. Gotcha. It can. That would maintain pre-3.3 compatibility.

@parkr does your implementation ensure the *source* doesn't exist or the *destination*? We definitely want the source checked, so that I can overwrite a theme file, but thinking we also want to check the destination, to prevent invisibly clobbering a site file.

benbalter: > this is a breaking change.

Why do you say that? Why does it have to be?

ashmaroli: > Can a page in assets/ not have a custom permalink in YAML?

it can.. any item in `assets` behaves like any other page or static-file..

ashmaroli: > Why do you say that? Why does it have to be?

That's my understanding.. because when jekyll builds the site, the stylesheet will be in `_site/assets/css` instead of `_site/css` 

--
/cc @parkr

ashmaroli:
@benbalter, I have a theme gem repo at [ashmaroli/solitude](https://github.com/ashmaroli/solitude) that uses this workflow..

_includes/head.html
ashmaroli: Is there a liquid tag to call the stylesheet? Like when using `jekyll-assets`?

pathawks: > it can.. any item in assets behaves like any other page or static-file..

So then the URL should not need to change. Let's keep it `/css/main.css`

benbalter: If we require Jekyll `~> 3.3`, I think we can make this a non-breaking change. There are two possibilities:

1. User overrode `head.html` to hard code `/css/main.css` and has a `/css/main.scss` already in their site (from `jekyll new`)

2. User has default head, will have a `/css/main.scss` file but it won't do anything. New head will use `/assets/style.css` which we will provide.

Having `/assets/style.scss` write to `/css/main.css` feels funny, especially because this is the default/starter site, and I suspect most users wouldn't know to look in `/assets` to replace a file in outputted to `/css` which is already confusing enough.

ashmaroli: > I suspect most users wouldn't know to look in /assets

Thats where jekyllrb plays an important role..

ashmaroli: Can either of you write a quick test for this case?

ashmaroli: I have manually tested  both scenarios and only hardcoding `assets/css/main.css` seems to be working..

ashmaroli: > New head will use /assets/style.css which we will provide.

How do we provide a new head apart from a "default" head??

ashmaroli: > If we require Jekyll ~> 3.3, I think we can make this a non-breaking change. 

By breaking change I was implying that this would cause trouble unless the user manually intervenes and changes their site's source code. Sorry, if the mainstream definition implies something else.
`require`ing ~>3.3  will not solve things as of now..

parkr: > @parkr does your implementation ensure the source doesn't exist or the destination? We definitely want the source checked, so that I can overwrite a theme file, but thinking we also want to check the destination, to prevent invisibly clobbering a site file.
  
@benbalter It only checks for `relative_path`, so `assets/css/main.scss` will be written only if a file `assets/css/main.scss` does NOT exist in the site source. We don't check destination.

parkr:
We also have the option to add a note to `spec.post_install_message` that describes the changes.

parkr:
> We also have the option to add a note to spec.post_install_message that describes the changes.

@ashmaroli Does that work for you?

minima.gemspec
```diff
@@ -19,6 +19,22 @@ Gem::Specification.new do |spec|
...
+----------------------------------------------
+Thank you for installing minima 2.0!
+
+Minima 2.0 comes with a breaking change that
+renders '<your-site>/css/main.scss' redundant.
+That file is now bundled with this gem as
+'<minima>/assets/css/main.scss'.
+
+More Information:
+https://github.com/jekyll/minima/issues/43
```
parkr: This is great! Should we add something to the README telling folks how they can use their custom CSS instead, namely by moving their `css/main.scss` to `assets/css/main.scss` in their site source? Would be good I think.

ashmaroli:
> Should we add something to the README telling folks how they can use their custom CSS

yes, the README will need an update as well.

_includes/head.html
benbalter: > `assets/css/main.scss`

Can we talk through a bit the pros and cons of `assets/css/main.scss` versus `/assets/main.scss`? I'm just worried we're over optimizing what is otherwise a simple theme.

parkr:
> yes, the README will need an update as well.

OK great, so let's update the README and set the URL in the postinstall message to point to the readme.

_includes/head.html
ashmaroli: I'm here, if that comment was intended for me..

ashmaroli: I'm advocating `assets/css/main.scss`, so the following is in context to that:
### pros
- this is a breaking change (to me, at least). I feel, its the right time to initiate the practice of having delegated folders. `css` for stylesheets, `js` for scripts, `fonts` for well.. fonts. 
- having this set right now will show users/developers that Jekyll 3.3 supports delegated folders. minima is not just a theme, its also a boilerplate / reference for future theme gems.

### cons
- having to create two folders to house one file, while preparing to override default settings.
- (or), in the case of existing sites, having to create `assets/` and move their `css/` to it, instead of simply renaming `css/ -> assets/`

README.md
```diff
@@ -22,18 +22,19 @@ And then execute:
...
 
-@import "minima";
-```
+### Customization
+
+To override the default structure and style of minima, simply create the concerned directory at the root of your site, copy the file you wish to customize to that directory, and then edit the file.
+e.g., to override the [`_includes/head.html `](_includes/head.html) file to specify a custom style path, create an `_includes` directory, copy `_includes/head.html` from minima gem folder to `<yoursite>/_includes` and start editing that file.
 
-You can choose to override the [`_includes/head.html `](_includes/head.html) file to specify a custom style path.
+`css/main.scss` has now moved to a new directory, `assets/` within the theme gem itself. To customize that file, you'll have to create an `assets` directory at the root of your site and copy the entire `css/` over to `<your-site>/assets/`.
```
parkr: Thanks! This is a bit confusing to me to read. What do you think about:

> The site's default CSS has now moved to a new place within the gem itself, `assets/css/main.scss`. This is why running `jekyll new` does not output any CSS to your new site. To override the default CSS, create a file **in your site source** at `assets/css/main.scss` and add your custom CSS. If you would like to start with the template file in this repository, copy [`assets/css/main.scss`](https://github.com/jekyll/minima/blob/master/assets/css/main.scss) to your site at `<your-site>/assets/css/main.scss` and edit away!

minima.gemspec
parkr: This can now be changed to `https://github.com/jekyll/minima#customization` 😄

README.md
ashmaroli: Thanks for the input, but I think it was too much info crammed into a few lines.. not good when a confused user checks this section for an answer. So I altered it _a bit_. 

> The site's default CSS has now moved to a new place within the gem itself, `assets/css/main.scss`. To **override the default CSS**, the file has to exist at your site source. Do either of the following:
- Create a new instance of `main.scss` at site source. 
  - Create a new directory `css` at `<your-site>/assets/`
  - Create a new file `main.scss` at `<your-site>/assets/css/`
  - Add the frontmatter dashes, and
  - Add `@import "minima";`, to `<your-site>/assets/main.scss`
  - Add your custom CSS.
- Download the file from this repo
  - Create a new directory `css` at `<your-site>/assets/`
  - Create  a new file `main.scss` at `<your-site>/assets/css/`
  - Copy the contents at [assets/css/main.scss]
(https://github.com/jekyll/minima/blob/master/assets/css/main.scss) onto the `main.scss` you just created, and edit away!
- Copy directly from Minima 2.0 gem
  - Go to your local minima gem installation directory ( run `bundle show minima` to get the path to it ).
  - Copy the `css/` folder from there into `<your-site>/assets/`
  - Change whatever values you want, inside `<your-site>/assets/css/main.scss`

parkr: OK. I'd like to release this today or tomorrow and this PR is the last piece.

ashmaroli:
> I'd like to release this today or tomorrow and this PR is the last piece.

The PR is ready for one final review.
Regarding the release, may I suggest that you release Jekyll 3.3 and Minima 2.0 at the same time, as our templates do not lock minima to a specific major_version.. ?

_includes/head.html
benbalter: @ashmaroli what's the value of delegated folders? Especially for a small boilerplate site? Anyone sophisticated enough to create a site with multiple javascript or css files will likely be sophisticated enough to not use the default theme. Put another way, if we do this, we have to assume every other theme does the same. Is that the precedence we want to set? 

@parkr can the new assets folder support support sub-folders?

parkr: @benbalter Yes! It supports sub-folders.

ashmaroli:
@jekyll/minima, @jekyll/core:  I'm placing this on vote.. till Parker decides to end it..
Pls exercise your vote on this comment itself:
`+1` for `assets/css/main.scss`
`-1` for `assets/main.scss`

Thanks

Strangehill:
the styles aren't working when I try this on localhost --- not sure whether that's to be expected, will it need 3.3 to work?

envygeeks:
From a Jekyll Assets standpoint assets/css/main.css would allow us to jimmy ourselves into Jekyll's place and serve all assets with some magic, making things a bit more magical.  But at the end of the day we can do that anyways really. So I don't care either way... just sayin'

parkr:
> the styles aren't working when I try this on localhost --- not sure whether that's to be expected, will it need 3.3 to work?

@Strangehill Yes, you need Jekyll 3.3. You can run the `master` branch and it will work.

ashmaroli:
I've decided that **Minima 2.0** shall release with `assets/main.scss` instead of `assets/css/main.scss`
It can be changed if required, in a future release.

parkr:
@jekyllbot: merge +major

davidyell:
This breaks all default Jekyll installs. I've just installed the latest gem and my site built without any css due to the theme looking in `assets/main.css` instead of `css/main.css`. I don't know if this is an issue with Jekyll, Octopress or this theme.

ashmaroli:
Do you have the source code online for us to debug?

ashmaroli:
The Issue here is that you **have to use** `Jekyll 3.3.0 and above` to read `css` in `/assets/`

davidyell:
@ashmaroli Great, thanks. 👍 

Might be worth adding that to the readme file for future users. I'd also be tempted to ask the guys over at https://github.com/octopress/octopress to link to this themes repo somewhere in their docs as it took me a while to find the repo.

ashmaroli:
What's puzzling is that Minima 2.0 **is locked to** Jekyll 3.3 n above.. How were you able to successfully run Minima 2.0 against Jekyll 3.2.1?? :thinking:

davidyell:
That I'm not sure of. I'm a Ruby noob!

ashmaroli:
@benbalter Minima 2.0.0 got released before #59 was merged. So, while the issue reported above wont occur with the next release, I think, its time we release a new version a.s.a.p..

envygeeks:
At the speed of sap.

parkr:
2.1.0 released.

envygeeks:
That wasn't even my intention so why are you trying to put words in my mouth and make that my intention? If you need to know I was poking fun at the word "a.s.a.p".  Not everybody is out to get you and making wild assumptions like what you just did is how most beefs start.

parkr:
> That wasn't even my intention so why are you trying to put words in my mouth and make that my intention? If you need to know I was poking fun at the word "a.s.a.p". Not everybody is out to get you and making wild assumptions like what you just did is how most beefs start.

@envygeeks I'm sorry – I believed that you were referring to how slowly we were releasing the new version, because we hadn't done it and #59 was merged a few weeks ago. I don't mean to put words in your mouth, but please understand that when you say something like that, it is perfectly reasonable that we will take offense to your words. The way your comment comes off can have just as much of an impact as what you actually write. A.S.A.P doesn't line up to "At the speed of sap" in terms of the initials so I didn't think you meant that. Sorry for the misunderstanding.

envygeeks:
It doesn't need to align, fully.  The joke is in what "a.s.a.p" implies and what actually happens most of the time.

DirtyF:
@parkr thanks for the release :shipit: , @envygeeks thanks for the laugh 😄

PR #44, opened 2016-09-24
Use standardized tab size of 2 spaces

zdroid:
Jekyll uses 2-spaces tab size for all its files (including the Sass in jekyll/jekyll `site/` directory), so they should be used here, too.

### DIFF
--- a/_sass/minima.scss
+++ b/_sass/minima.scss
@@ -24,15 +24,15 @@ $on-laptop:        800px !default;
 
 // Use media queries like this:
 // @include media-query($on-palm) {
-//     .wrapper {
-//         padding-right: $spacing-unit / 2;
-//         padding-left: $spacing-unit / 2;
-//     }
+//   .wrapper {
+//     padding-right: $spacing-unit / 2;
+//     padding-left: $spacing-unit / 2;
+//   }
 // }
 @mixin media-query($device) {
-    @media screen and (max-width: $device) {
-        @content;
-    }
+  @media screen and (max-width: $device) {
+    @content;
+  }
 }
 
 // Import partials.
--- a/_sass/minima/_base.scss
+++ b/_sass/minima/_base.scss
@@ -4,8 +4,8 @@
 body, h1, h2, h3, h4, h5, h6,
 p, blockquote, pre, hr,
 dl, dd, ol, ul, figure {
-    margin: 0;
-    padding: 0;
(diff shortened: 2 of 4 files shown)


DirtyF:
@jekyll/ecosystem @jekyll/core To avoid this, what about adding an `.editorconfig` to all official jekyll plugins and themes ?

envygeeks:
The idea was rejected a while back but that editorconfig was too personal.  From my perspective, I'm happy to accept one if it's pretty standard, 2 spaces for yml/ruby, no tabs, that's it, anything else is personal IMO.

zdroid:
@envygeeks This is the `.editorconfig` I stick with (probably the most popular one):
```ini
# editorconfig.org

root = true

[*]
charset = utf-8
end_of_line = lf
indent_size = 2
indent_style = space
insert_final_newline = true
trim_trailing_whitespace = true

[*.md]
trim_trailing_whitespace = false
```

I can't see anything personal there. We stick to 2 spaces everywhere, use LF endings, use UTF-8 charset, trim trailing whitespace (unless it's Markdown) and instert final newline at the end of file. Just the basic things to ensure it'll go alright.

envygeeks:
I'm cool with that that editorconfig, you're right nothing personal there it's all standard and what we enforce ourselves anyways (or should.) So if you wanna ship a pull to jekyll/ekyll and here for that I'll vote for it.

spudowiar:
I'd quite like a `script/fmt` (standardised for all official Jekyll projects) that lints HTML and SCSS (and other assets). `editorconfig` is all well and good but the linting tools are poor and un`editorconfig`ed code does make it into repositories by accident.

zdroid:
@spudowiar The possibly made script isn't auto-implemented in the modern code editors. Atom, Brackets and Sublime — the 3 most popular code editors in GitHub community — support that simple file `.editorconfig`. Editorconfig is here exactly to prevent such things to happen in future, the past mistakes have to be fixed manually.

spudowiar:
@zdroid I'm not saying to _not_ include an `editorconfig`, rather that this is something to complement an `editorconfig` :smile: I'm just too lazy to open another issue for it!

zdroid:
@spudowiar Oh, I completely agree. It'd ease the PR reviewing a lot.

parkr:
@jekyllbot: merge +dev

PR #46, opened 2016-09-24
priority to site.url when site.github.url is available

oncleben31:
# Issue:
* When using github-pages gem and creating a site with the command `bundle exec jekyll new . -force`, there is inconstancies in the url links after build.
* even if site.url or site.baseurl are set, the site.github.url was used in the header.html snippet
* when running the site locally there are link to github.com instead of local jekyll server IP

# Solution suggested:
* The patch is consistent with the previous commit: addf9a3787ef079fd26b8fd54d160f0cb684e57f
* The patch allows to generate only local links when using `bundle exec jekyll serve`

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -3,7 +3,7 @@
   <div class="wrapper">
 
     {% assign custom_url = site.url | append: site.baseurl %}
-    {% assign full_base_url = site.github.url | default: custom_url %}
+    {% assign full_base_url = custom_url | default: site.github.url %}
     <a class="site-title" href="{{ full_base_url }}/">{{ site.title | escape }}</a>
 
     <nav class="site-nav">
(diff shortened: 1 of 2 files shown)


ashmaroli:
Thank you for this! I missed this when I pushed the other commit. 
There is one other place where a similar patch may be made : `example/index.html`
You can make the change there also.

--
*side-note:* The code in `example/index.html` will be moved to a new `_layouts/home.html` very soon..

oncleben31:
@ashmaroli done
I tested it with a first run with
``` yaml
baseurl: "/minima"
```
and a second with
``` yaml
#baseurl: "/minima"
github:
  url: "github.com/minima"
```

All seems to work as planned ;-)

benbalter:
LGTM.

parkr:
@jekyllbot: merge +minor

PR #52, opened 2016-09-29
add `header_pages` config to link only specific files in header

### LINKED ISSUE #19: Ability to specify Pages in Header
It's awesome that the theme auto-includes a list of pages in the header, but once you get above a certain number of pages, it gets a bit unwieldy. 

It also be cool to be able to set the order of the pages. Thinking maybe some sort of array in the config of page paths to include, falling back to all pages if not set.

Crunch09:
fixes #19

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -4,17 +4,8 @@
 
     {% assign custom_url = site.url | append: site.baseurl %}
     {% assign full_base_url = custom_url | default: site.github.url %}
-    {% if site.header_pages %}
-      {% assign header_pages = "" | split:"|"  %}
-      {% for path in site.header_pages %}
-        {% assign header_page = site.pages | where: "path", path | first %}
-        {% if header_page %}
-          {% assign header_pages = header_pages | push: header_page %}
-        {% endif %}
-      {% endfor %}
-    {% else %}
-      {% assign header_pages = site.pages %}
-    {% endif %}
+    {% assign default_paths = site.pages | map: "path" %}
+    {% assign page_paths = site.header_pages | default: default_paths %}
     <a class="site-title" href="{{ full_base_url }}/">{{ site.title | escape }}</a>
 
     <nav class="site-nav">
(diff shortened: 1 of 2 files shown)


Crunch09:
cc: @benbalter

_includes/header.html
```diff
@@ -4,6 +4,17 @@
 
     {% assign custom_url = site.url | append: site.baseurl %}
     {% assign full_base_url = custom_url | default: site.github.url %}
+    {% if site.header_pages %}
+      {% assign header_pages = "" | split:"|"  %}
```
benbalter: What's going on here?

_includes/header.html
```diff
@@ -4,6 +4,17 @@
 
     {% assign custom_url = site.url | append: site.baseurl %}
     {% assign full_base_url = custom_url | default: site.github.url %}
+    {% if site.header_pages %}
```
benbalter: Could this be written as `{% assign header_pages = site.header_pages | default: site.pages %}`?

benbalter: I believe, if we do that, we'll only need to iterate through the pages once.

benbalter:
Nice start! A few questions on the implementation to try to DRY things up a bit, but I really like it.

_includes/header.html
Crunch09: I got this from https://talk.jekyllrb.com/t/how-do-you-add-items-to-an-array-in-jekyll/324/4 to initialize an empty array.

_includes/header.html
Crunch09: Good suggestion! Yes we could do that,
but i think we should use something like this then:
```
{% assign default_paths = site.pages | map: "path" %}
{% assign header_pages = site.header_pages | default: default_paths %}
```
To make sure in both cases we're dealing with `Strings` and not in one case `Strings`, in the other one with `Page` objects.

In the remaining loop we could then convert these back  to `Pages` one by one. What do you think?

Crunch09:
@benbalter Thanks for the review! Improved it now to use only one loop

scoskey:
Up vote.

benbalter:
Thanks @Crunch09! Sorry for the delay.

Crunch09:
@benbalter No worries! Thank you for merging it.

PR #54, opened 2016-10-05
Update gem regex to include new `assets` directory

ashmaroli:
Update the regex in `minima.gemspec` to include `assets` directory in **Minima-2.0** gem.
Should've covered this in #43, my bad.. 

--
/cc @benbalter, @parkr


parkr:
@jekyllbot: merge +bug

ashmaroli:
@parkr, this was not a bugfix.. rather a consequence of enhancement

Issue #56, opened 2016-10-06
Add `relative_url` filter to minima 


ashmaroli:
I think we should update relevant files to use the new `relative_url` as a live example of its usage. I'm not opening a PR as I'm not sure regarding the `site.github.url` bit.

--
/cc @jekyll/minima

benbalter:
> I'm not sure regarding the site.github.url bit.

GitHub Pages now sets `site.url` and `site.baseurl` in anticipation of Jekyll 3.3, meaning `absolute_url` and `relative_url` should always work as expected, even with no value explicitly set.

ashmaroli:
Thats great! If either of you haven't started work on a pull, regarding this, i will start..

benbalter:
> If either of you haven't started work on a pull, regarding this, i will start..

Please do!

benbalter:
> GitHub Pages now sets site.url and site.baseurl

The relevant PR, if you're interested: https://github.com/jekyll/github-metadata/pull/76

ashmaroli:
Thanks :smiley:

PR #57, opened 2016-10-06
Use  new `absolute_url` and `relative_url` filters in minima

ashmaroli:
replace existing liquid assignments `custom_url` and `full_base_url` with the new Jekyll filter `relative_url`
though **canonical URL** should always be an absolute URL (use `absolute_url` filter)

Ref: #56

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -7,7 +7,7 @@
   <meta name="description" content="{{ page.excerpt | default: site.description | strip_html | normalize_whitespace | truncate: 160 | escape }}">
 
   <link rel="stylesheet" href="{{ "/assets/main.css" | relative_url }}">
-  <link rel="canonical" href="{{ page.url | replace:'index.html','' | relative_url }}">
+  <link rel="canonical" href="{{ page.url | replace:'index.html','' | absolute_url }}">
   <link rel="alternate" type="application/rss+xml" title="{{ site.title | escape }}" href="{{ "/feed.xml" | relative_url }}">
   
   {% if jekyll.environment == 'production' and site.google_analytics %}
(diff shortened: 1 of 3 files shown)


_includes/head.html
```diff
@@ -6,11 +6,9 @@
   <title>{% if page.title %}{{ page.title | escape }}{% else %}{{ site.title | escape }}{% endif %}</title>
   <meta name="description" content="{{ page.excerpt | default: site.description | strip_html | normalize_whitespace | truncate: 160 | escape }}">
 
-  {% assign custom_url = site.url | append: site.baseurl %}
-  {% assign full_base_url = custom_url | default: site.github.url %}
-  <link rel="stylesheet" href="{{ "/assets/main.css" | prepend: full_base_url }}">
-  <link rel="canonical" href="{{ page.url | replace:'index.html','' | prepend: full_base_url }}">
-  <link rel="alternate" type="application/rss+xml" title="{{ site.title | escape }}" href="{{ "/feed.xml" | prepend: full_base_url }}">
+  <link rel="stylesheet" href="{{ "/assets/main.css" | relative_url }}">
+  <link rel="canonical" href="{{ page.url | replace:'index.html','' | relative_url }}">
```
pathawks: It looks like `canonical` used to be an absolute URL, and it _really needs to be_ an absolute URL.

I agree with you that `stylesheet` is better off as a relative URL. I could go either way with the feed link.

ashmaroli: I agree with you that a **canonical url** *should* be an absolute URL. My bad.. will fixup right away..

ashmaroli:
> We'll want to update the gemspec to require >= 3.3.

I'll start a PR for that as well..

benbalter:
@jekyllbot: merge +minor

benbalter:
Thanks, once again, @ashmaroli! 😄

PR #60, opened 2016-10-06
Basic i18n implementation

oncleben31:
So I've redo my try to put some basic i18n. This time it's more clean and I use yaml files to store the locales strings.

It still work in progress as there are two points on my todo list:
-   (mandatory) ~~make it work with the locales files in the theme folder not in the example/_data folder~~ (**Edited** not possible currently)
-   find an easy way to localize the dates. (**Edited** done)

cc/ @ashmaroli Have you any comment on that ?

### DIFF
--- a/_includes/translated_date.html
+++ b/_includes/translated_date.html
@@ -1,47 +1,57 @@
 {% comment %}
 This include allows to tranlate the days and months in the date by using strings in _data/locales/[lang].yml
-This include can be called with two parameters:
+This include can be called with three parameters:
  - include.date: the date to be processed
  - include.format: the date format. (If empty a default format will be used)
+ - include.lang: lang to use for translation.
 {% endcomment %}
 
 {% comment %}
-If the parameter date format is empty the default format the one set in _config.yml or at least "%b %-d, %Y" is used
+If the parameter date format is empty the default format set in _config.yml is used.
+If there is no default in _config, "%b %-d, %Y" is used.
 {% endcomment %}
+
 {% assign date_format_to_be_translated = include.format | default: site.date_format | default: "%b %-d, %Y" %}
 
 {% comment %}
 Init to have the indice of the day in the week (num_day) and the indice of the month in the year (num_mont)
+Init the lang parameter depending if it's set in the include call, in the page or in the site settings
 {% endcomment %}
+
 {% assign num_day = include.date | date: "%w" | plus: 0 %}
 {% assign num_month = include.date | date: "%-m" | plus: 0 %}
+{% assign lang = include.lang | default: page.lang | default: site.lang %}
 
 {% comment %}
 Use tranlated abbreviated weekday if "%a" is used in the format and the translation is available in _data/locales/[lang].yml
 {% endcomment %}
-{% if site.data.locales[site.lang].abbreviated_weekday[num_day] %}
-  {% assign date_format_to_be_translated = date_format_to_be_translated | replace: "%a",  site.data.locales[site.lang].abbreviated_weekday[num_day] %}
+
+{% if site.data.locales[lang].abbreviated_weekday[num_day] %}
+  {% assign date_format_to_be_translated = date_format_to_be_translated | replace: "%a",  site.data.locales[lang].abbreviated_weekday[num_day] %}
 {% endif %}
 
 {% comment %}
 Use tranlated full weekday if "%A" is used in the format and the translation is available in _data/locales/[lang].yml
 {% endcomment %}
-{% if site.data.locales[site.lang].full_weekday[num_day] %}
-  {% assign date_format_to_be_translated = date_format_to_be_translated | replace: "%A",  site.data.locales[site.lang].full_weekday[num_day] %}
+
+{% if site.data.locales[lang].full_weekday[num_day] %}
+  {% assign date_format_to_be_translated = date_format_to_be_translated | replace: "%A",  site.data.locales[lang].full_weekday[num_day] %}
 {% endif %}
 
 {% comment %}
 Use tranlated abbreviated month if "%b" is used in the format and the translation is available in _data/locales/[lang].yml
 {% endcomment %}
-{% if site.data.locales[site.lang].abbreviated_month[num_month] %}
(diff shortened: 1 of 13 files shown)


ashmaroli:
Lovely, I dig the use of `_data/locales/[lang].yml`
There are two ways to consider..

--
When setting `default: [params]` within the liquid tag (inside `_layouts/home.html`) for every i18n string,
you dont need to explicitly set the `site.lang: en` in `_config.yml` and  you can comment that out as well.

```yaml
# Basic i18n 
# uncomment the line below to have minima
# use the data in _data/locales/[lang].yml 
#
# lang: fr

```
and hence wont *explicitly* need a `en.yml`

--
The second approach (**preffered but not possible as of now**) is to simply remove the default filters in the liquid tags for i18n strings:
```html
<h1 class="page-heading">{{ site.data.locales[site.lang].posts }}</h1>

<p class="rss-subscribe">{{ site.data.locales[site.lang].subscribe }}
  <a href="{{ "/feed.xml" | relative_url }}">{{ site.data.locales[site.lang].via_RSS }}</a>
</p>
```
But the problem with this is Jekyll doesnt read data files from within the gem as of v3.3 so we cant ship default locale data alongwith the gem, and the user will have those strings missing from their site till he manually adds
So this way is closed for now.

ashmaroli:
Overall, this PR LGTM so far except for that small nit I mentioned above.. 
please proceed..

Also, please change to a more descriptive title.. `(2nd try)` is not required.
*for example:*
**`Add basic i18n support`** or *similar*

oncleben31:
Thank you for you very constructive comments. I will work on it next week.

oncleben31:
@ashmaroli any recommandation for the location to store the [lang].yml files ?
At the end the user will put it in the `_data` folder. But on a fresh install the folder will be empty and the user should find a message somewhere (documentation or default `_config.yml`file or README on Github, ....) explaining how to grab an existing one or create this file and where to put it.

My first opinion was on `_data` folder in minima theme folder, but you just explain me Jekyll currently can't access it. So it's perhaps not the best option.

ashmaroli:
Sorry, I dont have a solid recommendation as of now..

ashmaroli:
One way is to store the data files in this repo itself.. like in your PR.. as `example/_data/fr.yml`. Add a section in README informing Users wanting i18n to have to copy from that folder

oncleben31:
My PR is ready for review with strings and dates translation.

/cc @ashmaroli

_includes/translated_date.html
```diff
@@ -0,0 +1,20 @@
...
+
+{% if date_format_to_be_translated contains "%b" %}
+  {% assign replacement =  site.data.locales[site.lang].abbreviated_month[num_month] | default: "%b" %}
+  {% assign date_format_to_be_translated = date_format_to_be_translated | replace: "%b",  replacement %}
+{% elsif date_format_to_be_translated contains "%B" %}
+  {%assign replacement =  site.data.locales[site.lang].full_month[num_month] | default: "%B" %}
+  {% assign date_format_to_be_translated = date_format_to_be_translated | replace: "%B",  replacement %}
+{% endif %}
+
+{{ include.date | date: date_format_to_be_translated }}
```
ashmaroli: Whoa! that's a **LOT** of assignments.. IMHO, there has to be a better way to have this implemented..

I advice you use comments to provide embedded documentation for this. Helps others who read your code..:
- `{% comment %}{% endcomment %}` for liquid comments that do not get carried over to final site markup, and
- `<!-- -->` for attribution (the source of this code, for example, with licence info, if available.) that does get carried over to site markup.

example/_config.yml
```diff
@@ -9,6 +9,15 @@ baseurl: "/minima"
 twitter_username: jekyllrb
 github_username:  jekyll
 
+### Basic localisation (l10n)
+# If you customize the lang parameter, minima will try to use the strings in _data/locales/[lang].yml
+# Read https://github.com/jekyll/minima/README.md for more information.
+#lang: fr
```
ashmaroli: I prefer comments to **always** have a whitespace after the `#`. Looks better.  
(Plus, it doesnt seem like a reference-tag.. 
like how we would link to a heading in HTML e.g. `/about.html#work-experience` though I agree, its of no significance here.)

example/_config.yml
```diff
@@ -9,6 +9,15 @@ baseurl: "/minima"
...
+
+# Even the day and month names will be translated. You can customize your date by
+# uncommenting the following line and customizing the date format:
+#date_format: "Le %-d %B %Y"
```
ashmaroli: and here.. :wink:

example/_data/locales/fr.yml
```diff
@@ -0,0 +1,47 @@
...
+  - avr.
+  - mai
+  - juin
+  - juil.
+  - août
+  - sept.
+  - oct.
+  - nov.
+  - déc.
+full_month:
```
ashmaroli: why use lowercase names suddenly? If `en.yml` has `Jan.` | `Feb.`, `fr.yml` should also be `Jan.` | `Fev.` etc

ashmaroli:
Not a big fan of having an `include` with **LOTS** of assignments to implement date-translation..
- we should have a built-in system to have that done in a cleaner and simpler way..

If, `ThemeDataReader` doesnt get merged `upstream`, and if a better alternative doesn't surface, 
I'll release that monkey-patch as a separate gem..

example/_data/locales/fr.yml
oncleben31: FYI in french the stadard format for dates use months in lowercase. Like "Mercredi 12 octobre 2016"

ashmaroli: :smiley: Yes, I'd struck that comment out when I learnt about this convention..

oncleben31:
@ashmaroli your comments have been taken into account.
I've succeeded to reduced the number of `assign`. I probably could do more but the code readability will be really low.

ashmaroli:
Thank you very much @oncleben31, I would still additionally add a new line between every `{% endcomment %}` and `{% assign [..] %}`.. but you may leave it as is for now..

:+1:  for the effort put in this PR..

--
/cc @jekyll/minima  Any comments?

DirtyF:
Closing this, as we don't intend to support i18n in this theme. Thanks for your effort @oncleben31

PR #61, opened 2016-10-07
Example should require Jekyll 3.3

DirtyF:
As we ship assets, I assume Jekyll 3.3 is required.

/cc @jekyll/minima


parkr:
@jekyllbot: merge +dev

Issue #69, opened 2016-10-27
Configuring time format


ward:
I notice that the time format in _layouts/home.html and _layouts/post.html is hardcoded. Is it possible/would it make sense to make this configurable in _config.yml?

Am I correct that right now my main option to change the format is to make a copy of those files and edit those copies?

DirtyF:
@ward you're correct about the copy part. 

Are you interested in submitting a PR to handle date format in configuration file?

ward:
Sure, I can take a look at it. I should have some time for it over the weekend.

ward:
Well I looked at it on the tram home and it seemed simple enough. https://github.com/jekyll/minima/pull/70

PR #70, opened 2016-10-28
Use date_format if set in configuration

ward:
If it is not set, fall back to the previous default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,20 @@ The site's default CSS has now moved to a new place within the gem itself, [`ass
 
 --
 
+### Change default date format
+
+You can change the default date format by specifying `site.minima.date_format`
+in `_config.yml`.
+
+```
+# Minima date format
+# refer to http://shopify.github.io/liquid/filters/date/ if you want to customize this
+minima:
+  date_format: "%b %-d, %Y"
+```
+
+--
+
 ### Enabling comments (via Disqus)
 
 Optionally, if you have a Disqus account, you can tell Jekyll to use it to show a comments section below each post.
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -11,11 +11,8 @@ layout: default
   <ul class="post-list">
     {% for post in site.posts %}
       <li>
-        {% if site.date_format %}
-          <span class="post-meta">{{ post.date | date: site.date_format }}</span>
-        {% else %}
-          <span class="post-meta">{{ post.date | date: "%b %-d, %Y" }}</span>
-        {% endif %}
+        {% assign date_format = site.minima.date_format | default: "%b %-d, %Y" %}
+        <span class="post-meta">{{ post.date | date: date_format }}</span>
 
         <h2>
           <a class="post-link" href="{{ post.url | relative_url }}">{{ post.title | escape }}</a>
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -11,7 +11,11 @@ layout: default
   <ul class="post-list">
     {% for post in site.posts %}
       <li>
-        <span class="post-meta">{{ post.date | date: "%b %-d, %Y" }}</span>
+        {% if site.date_format %}
+          <span class="post-meta">{{ post.date | date: site.date_format }}</span>
+        {% else %}
+          <span class="post-meta">{{ post.date | date: "%b %-d, %Y" }}</span>
+        {% endif %}
 
         <h2>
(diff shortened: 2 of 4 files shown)


ward:
See issue https://github.com/jekyll/minima/issues/69

DirtyF:
@ward LGTM, would you mind adding an example section in the README file?

parkr:
I might do `site.minima.date_format`. Starting to get more concerned about overlapping parameters.

ashmaroli:
> I might do site.minima.date_format. Starting to get more concerned about overlapping parameters.

I have written a plugin [`jekyll-data`](https://github.com/ashmaroli/jekyll-data) that reads data files in theme gems and also provides a `theme` liquid-drop as an alias for site.data.[theme-name].
If @jekyll/minima is fine with `require`-ing that plugin, then we can ship a `minima.yml` inside 
`<minima-root>/_data` that contains *minima*-specific configurations.

```html
<!-- 'theme.date_format' is an alias for 'site.data.minima.date_format' -->
{% assign date_format = theme.date_format %}
{% if site.date_format %}
  <span class="post-meta">{{ post.date | date: site.date_format }}</span>
{% else %}
  <span class="post-meta">{{ post.date | date: "%b %-d, %Y" }}</span>
{% endif %}
```

oncleben31:
Don't know if there is any performance impact. I though that using the `default` liquid filter is better than the `if ... else ... endif`

Something like
```
{% assign date_format = theme.date_format | default:  "%b %-d, %Y" %}
<span class="post-meta">{{ post.date | date: date_format }}
````

Is there any pros and cons?

**Comment by ward** [comment]
* Documentation: Adding a line in customization seems like it should be enough, correct?
* Naming: I tried a quick search before deciding date_format to see if there was some jekyll default for this, but did not find anything. Found that quite surprising actually, I imagine plenty of people would want to change this, regardless of what theme they use. (Though I admit on older jekyll versions I did it by editing the theme files directly)
* `default` vs `if`: I did (and do) not know enough about liquid. If `default` is more performant, then it is the preferred choice indeed.

**Comment by DirtyF** [comment]
@ward  

yeah, based on what Parkr recommends, I was thinking of something like:

## Change default date format

You can do so by specifying `theme.minima.dateformat` in `_config.yml`:

# Minima date format
# refer to http://shopify.github.io/liquid/filters/date/ if you want to customize this
site:
  minima:
    date_format: "%b %-d, %Y"

I would even add the above snippet by default  in the `_config.yml` and use it right away in the theme, if you ask me, so you can simplify the layout.

**Comment by benbalter** [comment]
If it's a site configuration flag, I don't understand why it'd be in a data file (content), rather than `_config.yml` (site configuration)

**Comment by ashmaroli** [comment]
> If it's a site configuration flag, I don't understand why it'd be in a data file (content), rather than _config.yml (site configuration)

because *to me*, `site.minima.myvariable` is not site-configuration, but rather, a theme-configuration; and jekyll-data plugin expects to find theme configuration in `/_data` dir.

I may be wrong in thinking that data files can blur the lines between content and configuration, but its worth a shot. :smiley:

**Comment by DirtyF** [comment]
@ward Thanks!

**Comment by ward** [comment]
Any update on when this will be merged in?

**Comment by ashmaroli** [comment]
FWIW, the most recent version of `jekyll-data` plugin reads a `_config.yml` within theme-gem. This would allow theme developers to ship theme specific default configuration within the theme-gem..
```yaml
# <minima-gem>/_config.yml
#
# ** currently works only when using `jekyll-data` plugin` **
#
# minima-specific DEFAULTS (currently set using the `default` liquid-filter)
# will be specified below and accessed via `{{ site.minima.myvariable }}`
#
minima:
  lang: "en"
  date_format: "%b %-d, %Y"
```

benbalter:
@jekyllbot: merge +minor.

Thanks @ward! Sorry for the delay in getting this merged.

Sathors:
Would it be possible to have a release that includes this change?

Thanks.

DirtyF:
@ward @Sathors FYI if you want to benefit from the latest changes on the theme you can always point to the `master`branch of this repo meanwhile.

```yaml
gem "minima", :github => "jekyll/minima", :branch => "master"
```

Sathors:
Thanks for the tips, I knew about it, but it is more maintenance burden.

More than anything, it is because the Readme does indicate this option, so one is expecting it to be present in the release.

Thank you nonetheless.

PR #76, opened 2016-11-08
Demo minima on GitHub Pages

### LINKED ISSUE #67: Demo
Thank you so much for this great theme!

How about a demo for the always latest version of the theme?

DirtyF:
fix #67

/cc @jekyll/minima


pathawks:
This would not allow the example site to showcase the current version of Minima, but would rely on the version included in GitHub Pages, is that correct?

benbalter:
> This would not allow the example site to showcase the current version of Minima, but would rely on the version included in GitHub Pages, is that correct?

Correct. I'm fine with it essentially showing the released version, unless there's any objection.

ashmaroli:
I'm :-1: on this primarily because:
- IMHO, a Jekyll theme-repo root should only contain directories relevant to the theme, **plus** a directory named either ***example*** or ***demo*** which will contain files and directories generated by the latest version of Jekyll.
- The separation of theme-gem files and source-files will help us **showcase** the actual working relationships between a theme-gem-dir and source-dir:
  - `_layouts`, `_includes`, `_sass` & `assets` may not be present inside the `source_dir` to be read by _Jekyll_
  - The files inside directories above, if present inside same directories at `source_dir`, will override theme templates. (currently not demo-ed, though)

benbalter:
@DirtyF Thanks for implementing this and sorry it took so long to review.

In order to facilitate demos, local development, and legacy theme workflows (fork and edit), Jekyll themes were intended to be just Jekyll sites that had a Gemspec added. I really like the simplicity this PR allows for and how much it simplifies the local development experience. Anyone who knows how to edit a Jekyll site can now contribute to minima. Thank you!

@jekyllbot: merge +dev.

ashmaroli:
@benbalter Please add the preview link on the repo's index page

alxn:
@ashmaroli I guess like the rest of the [themes](https://pages.github.com/themes/):

> XXXX is a Jekyll theme for GitHub Pages. You can preview the theme to see what it looks like, or even use it today.

ashmaroli:
@alxn I agree with your point. But I also liked how earlier **I could make the distinction** between a **Jekyll Site** and the **Minima gem contents** while developing...

ashmaroli:
To add to my point above, say, I add a `_data` folder to the repo. Earlier I'd add it to the `example/` and my users would know its not included in the gem.. Now they'll have to inspect the `gemspec` and interpret the `regex` to do the same..

Issue #77, opened 2016-11-09
Use different syntax highlighting style


lunaryorn:
How do I change the syntax highlighting style when using minima as a theme gem?

I tried to define my own `_sass/minima/syntax-highlighting.scss` but it doesn't seem to get picked up by Jekyll.  I could define my own stylesheet to `@import` in `assets/main.scss` but then I'd have to undo all of minima's syntax highlighting theme, which quickly gets tedious.

What's the recommended way here?

ashmaroli:
Rather very simple actually:
  - **copy** `<minima>/_sass/minima/_syntax-highlighting.scss` to your **source.root**
  - **edit** `<source>/_sass/minima/_syntax-highlighting.scss` and save.
  - `bundle exec jekyll serve`

lunaryorn:
@ashmaroli I'm sorry but as I said that did not work for me.  I have `_sass/minima/_syntax-highlighting.scss` with the following contents (just to make sure it works) but it's not picked up:

```
body {
  color: red;
}
```

I don't even get a syntax error if I put invalid SCSS into that file—Jekyll seems to entirely ignore the file.

ashmaroli:
do you have the source code online?

lunaryorn:
Forgot to say, that other partials from `_sass` are picked up just fine.  For instance, I've got `_sass/solarized/_palette.scss` with solarized color definitions, and I can perfectly import and use it via `@import "solarized/palette.scss"` from `assets/main.scss`.

lunaryorn:
@ashmaroli I just pushed my attempt to https://github.com/lunaryorn/lunaryorn.github.io/tree/try-override-minima-syntax-highlighting

lunaryorn:
@ashmaroli See https://github.com/lunaryorn/lunaryorn.github.io/commit/e8fe2aafcbdd0248a8d6967035eb7713fede3cf9.

The site has no special deps, just standard github pages.  You can build it with:

```
$ bundle install
$ bundle exec jekyll serve
```

ashmaroli:
okay.. it seems like the bug is actually in our documentation.. 
since our `main.scss` actually `@imports "minima";` which is within the theme-gem, our overrides are not being detected.. unless our `_sass` dir **at source** also has a `minima.scss`

ashmaroli:
## workaround
When attempting to override minima's partials, include a **`minima.scss`** inside the **`_sass` dir at source** as well.

lunaryorn:
[...]

PR #79, opened 2016-11-09
Update 'Customization' section in README

ashmaroli:
Update Customization section in README
Ref: Discussion in #77 

/cc @DirtyF


pathawks:
This seems like something that should be fixed with the way Jekyll handles themes.

ashmaroli:
> This seems like something that should be fixed with the way Jekyll handles themes.

more specifically **gem-based themes**
/cc @parkr

README.md
```diff
@@ -46,6 +46,38 @@ The site's default CSS has now moved to a new place within the gem itself, [`ass
...
 --
 
+When you override only a minima-sass-partial, it is not automatically imported because we're still importing the `minima.scss` within the theme-gem and that subsequently imports the partials with respect to itself, i.e. partials within the gem. Hence you should either include a *copy of `minima.scss` from the gem* inside the `_sass` directory at source or the overriding `/assets/main.scss` file should explicitly import the edited partial. :
+  e.g. To have an **edited** `_syntax-highlighting.scss` be rendered, you should either have
+
+```sass
+/* <your-site>/assets/main.scss */
+
+@import "minima";
+@import "minima/syntax-highlighting";
```
adamvoss: This seems like the right way to do it. I'm not sure we need to present the other way.

ashmaroli: Perhaps I wasn't clear enough, but if you read again, you'll see that the two contexts are different.
The first scenario is **`importing`** the partials to the SCSS file in `assets/` while the second scenario is having a copy of the `_sass/` directory with necessary partials.

The common factor in both scenarios or, the takeaway from this entire section is that:

> ***To override `minima/syntax-highlighting.scss` partial, one will need to pull `minima.scss` from their `<source>` either by importing it again, or by having it physically present.***

ashmaroli: Actually, a recent addition to the official documentation touches upon this aspect of customization

adamvoss: Maybe I am not understanding.  I read it as you you are presenting two different ways to accomplish the goal of overriding a minima-sass-partial.  One way you duplicate a file from the gem, the other you explicitly import your modifications.  The later way does not require the user to maintain a copy of _minima.scss_ making it more robust and less likely to break.  I was suggesting that be the approach that is presented to users and that the other one perhaps should not be.

Context: I encountered this exact issue first-hand with [json-schema.org](https://github.com/json-schema-org/json-schema-org.github.io).  On my own I figured out your first way of including a copy of `minima.scss`.  This worked, but then building the site broke when GitHub updated minima to v2.1.1 and `minima.scss` needed up be updated.  This would not have happened if I knew to just import the modified scss file.

PR #81, opened 2016-11-15
modify Rakefile to watch assets folder as well

ashmaroli:
This pull modifies the `Rakefile` to have the `assets` dir _watched_ during **development**

/cc @jekyll/minima

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -47,6 +47,7 @@ task :preview do
     base.join("_includes"),
     base.join("_layouts"),
     base.join("_sass"),
+    base.join("assets"),
     options["source"],
     :ignore => listen_ignore_paths(base, options),
     :force_polling => options['force_polling'],


benbalter:
Dumb question... why can't we just use `jekyll serve --watch`?

ashmaroli:
> why can't we just use `jekyll serve --watch`?

@benbalter, because:
- `--watch` is enabled by default for the `source_dir` (`/example` in this case) and not the `theme_gem (minima)` files.
- we **watch** `minima` files during development by having the Rakefile [define](https://github.com/jekyll/minima/blob/master/Rakefile#L46-L54) _what to watch_.
Hence `bundle exec rake preview` uses current **`minima`-branch** instead of an installed version.

benbalter:
Can we simplify that by collapsing the example files into the root?

ashmaroli:
> Can we simplify that by collapsing the example files into the root?

I'm against that because I'd like the root to only show directories actually relevant to the theme. We can easily test n showcase Jekyll reading extra-`source_dir` directories n files. Hope that's clear..

But if the other members on @jekyll/minima are fine with the move, I wont mind..

DirtyF:
@benbalter done in https://github.com/jekyll/minima/pull/76

benbalter:
I'd rather go with #76 here. Less code is more better. Thanks @ashmaroli, as always for the PR and sorry for the slow response.

PR #84, opened 2016-11-16
option to anonymize IP in Google Analytics

### LINKED ISSUE #83: Add option to anonymize IP in Google Analytics
Google provides the option to [anonymize the IP](https://support.google.com/analytics/answer/2763052?hl=en&ref_topic=2919631) of tracked user. This is required in different countrys to allow a legal tracking (e.g. in germany).
It would be nice, if this feature could be optional in minima.

    ga('set', 'anonymizeIp', true);

DirtyF:
Fix #83 

This PR just want to provide an easy way to allow site owners to disable IP tracking by Google, as it is requested in somes countries.

/cc @jekyll/minima

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ theme: minima
 And then execute:
 
     $ bundle
-    
+
 
 ## Usage
 
@@ -67,13 +67,17 @@ If you don't want to display comments for a particular post you can disable them
 
 ### Enabling Google Analytics
 
-To enable Google Anaytics, add the following lines to your Jekyll site:
+To enable Google Anaytics tracking, add or uncomment the following lines in your `_config.yml` and copy-paste your own tracking id:
 
 ```yaml
-  google_analytics: UA-NNNNNNNN-N
+  google_analytics:
+    tracking_id: UA-NNNNNNNN-N
+    anonymize_ip: false
 ```
 
-Google Analytics will only appear in production, i.e., `JEKYLL_ENV=production`
+Optionnally, to comply with your country own privacy policies, just set `anonymize_ip` to `true`.
+
+Google Analytics will only be active in productionn, i.e. when `JEKYLL_ENV=production`.
 
 ## Contributing
 
--- a/_includes/google-analytics.html
+++ b/_includes/google-analytics.html
@@ -4,8 +4,10 @@
   m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
   })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
-  ga('create', '{{ site.google_analytics }}', 'auto');
+  ga('create', '{{ site.google_analytics.tracking_id }}', 'auto');
+  {% if site.google_analytics.anonymize_ip == true %}
+    ga('set', 'anonymizeIp', true);
+  {% endif %}
   ga('send', 'pageview');
 
 </script>
-  
--- a/example/_config.yml
+++ b/example/_config.yml
@@ -9,9 +9,19 @@ baseurl: "/minima"
 twitter_username: jekyllrb
 github_username:  jekyll
 
+
 # Build settings
 markdown: kramdown
 theme: minima
+
+# theme settings
+# disqus:
+#   shortname: my_disqus_shortname
+# google_analytics:
+#   tracking_id: UA-NNNNNNNN-N
+#   anonymize_ip: false
+
+
 exclude:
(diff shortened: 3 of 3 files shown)


_includes/google-analytics.html
```diff
@@ -4,8 +4,10 @@
   m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
   })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
-  ga('create', '{{ site.google_analytics }}', 'auto');
+  ga('create', '{{ site.google_analytics.tracking_id }}', 'auto');
+  {% if site.google_analytics.anonymize_ip = true %}
```
ashmaroli: this should be `{% if site.google_analytics.anonymize_ip == true %}` (double equality)

DirtyF: fixed, thanks @ashmaroli

ashmaroli:
you'll have to update the section in README too..

README.md
```diff
@@ -67,13 +67,17 @@ If you don't want to display comments for a particular post you can disable them
 
 ### Enabling Google Analytics
 
-To enable Google Anaytics, add the following lines to your Jekyll site:
+To enable Google Anaytics tracking, add or uncomment the following lines in your `_config.yml` and copy-paste your own tracking id:
```
ashmaroli: uncomment?

README.md
```diff
@@ -67,13 +67,17 @@ If you don't want to display comments for a particular post you can disable them
...
 
 ```yaml
-  google_analytics: UA-NNNNNNNN-N
+  google_analytics:
+    tracking_id: UA-NNNNNNNN-N
+    anonymize_ip: false
 ```
 
-Google Analytics will only appear in production, i.e., `JEKYLL_ENV=production`
+Optionnally, to comply with your country own privacy policies, just set `anonymize_ip` to `true`.
```
ashmaroli: **typos:**
```diff
- Optionnally, to comply with your country own privacy policies, just set `anonymize_ip` to `true`.
+ optionally, to comply with your country's own privacy policies, just set `anonymize_ip` to `true`.
```

parkr:
People will want other options. What if we give a way to pass any arbitrary configuration options? Could they be easily passed as JSON?

DirtyF:
@parkr I have no idea about what Google Analytics allow, maybe @philipwalton could suggest us a way to pass configuration options via JSON?

_includes/google-analytics.html
```diff
@@ -4,8 +4,10 @@
   m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
   })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
-  ga('create', '{{ site.google_analytics }}', 'auto');
+  ga('create', '{{ site.google_analytics.tracking_id }}', 'auto');
+  {% if site.google_analytics.anonymize_ip == true %}
+    ga('set', 'anonymizeIp', true);
+  {% endif %}
```
parkr: I envision something like:

```liquid
{% for property in site.google_analytics.settings %}
  ga('set', {{ property[0] | jsonify }}, {{ property[1] | jsonify }});
{% endfor %}

File: README.md (hunk 2 above)
Comment (DirtyF) [inline README.md]: @ashmaroli I added example values in the provided `_config.yml`, they are commented by default. We could add them also in the default `_config.yml` shipped with Jekyll.

File: _includes/google-analytics.html (hunk 4 above)
Comment (philipwalton) [inline _includes/google-analytics.html]: @DirtyF @parkr that should work for most fields.  However, [some fields](https://devsite.googleplex.com/analytics/devguides/collection/analyticsjs/field-reference#create) can **only** be set in the `create` command, and some fields can **not** be set in the `create` command (they have to be set using the `set` command.

In other words, the following config wouldn't work with the above suggestion:

```yaml
site_speed_sample_rate: 10
anonymize_ip: true
transport: beacon
```

pathawks:
Maybe this is complicated enough that users who want to config specific options should just provide their own include file?

_includes/google-analytics.html
DirtyF: @philipwalton thanks for the detailed answer 😄 

@parkr So using your code snippet would imply asking users to check [analytics field references](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference) to see what properties they can set. A bit harsh for the majority of users don"t you think?

PR #85, opened 2016-11-16
fix hamburger menu for iOs

### LINKED ISSUE #80: Hamburger menu doesn't work on iOS
I'm using minima for http://www.lunaryorn.com and just noticed that the hamburger menu doesn't work on iPhone/iOS: When I tap the icon the menu doesn't pop up.   I went through a couple of phones on browerstack and could reproduce the problem with Safari on iPhones, whereas the menu works fine on Android devices with Google Chrome.  Note the issues does also not appear in the "responsive" design modes of Safari and Chrome on desktop.

I'm not so familiar with CSS, and don't really understand how the menu's implemented in Minima, so I'm at a loss.  I don't even know how to debug the issue given that it only occurs on the real phone.

Would really appreciate any help :)

melaniewalsh:
Adding a dummy onclick handler `onclick = "void(0)"` into the span element of the menu icon will make iOS recognize it as a clickable element. You won't be able to unclick the drop-down menu, however.

Fix #80

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -5,7 +5,7 @@
     <a class="site-title" href="{{ "/" | relative_url}}">{{ site.title | escape }}</a>
 
     <nav class="site-nav">
-      <span class="menu-icon">
+      <span class="menu-icon" onclick="void(0)">
         <svg viewBox="0 0 18 15" width="18px" height="15px">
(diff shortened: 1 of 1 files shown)


ashmaroli:
@melaniewalsh Thank you for digging out a solution. 
Have you tried using the [`pointer-events`](https://css-tricks.com/almanac/properties/p/pointer-events/) css to see if that would work too? 
I'm curious about that, and have no way to check it myself..

BasThomas:
What is the status on this?

Ambrosiani:
https://github.com/jekyll/minima/pull/111 is a better fix I think.

benbalter:
Thanks @melaniewalsh! Going to go with the solution in #111 instead, but appreciate the pull request. Keep 'em coming.

PR #86, opened 2016-11-26
Provide a hook for theme users to inject custom CSS

britter:
There is currently no easy way for users to override CSS rules defined by minima's main.css. In order to do this, one has to copy _includes/head.html to the user project for adding an additional stylesheet entry below the main.css.

This PR adds an empty custom.css and corresponding stylesheet declaration to minima. This way users can simply add assets/custom.css to their project if they need to override/add css rules.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -7,9 +7,10 @@
   <meta name="description" content="{{ page.excerpt | default: site.description | strip_html | normalize_whitespace | truncate: 160 | escape }}">
 
   <link rel="stylesheet" href="{{ "/assets/main.css" | relative_url }}">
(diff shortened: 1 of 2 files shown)


ashmaroli:
At a time when unnecessary HTTP requests are being actively minimized, why force *every* user to download an *empty stylesheet* ?

ashmaroli:
instead, I'd prefer that you add an empty partial in the `_sass` directory:
```sass
// Import partials.
@import
  "minima/base",
  "minima/layout",
  "minima/syntax-highlighting",
  "minima/custom"
;
```

benbalter:
I've proposed an alternative approach over in #117.

PR #88, opened 2016-11-29
handling content outside of posts

dret:
it doesn't make a lot of sense to pull content from `layout: home` pages under the *Posts* heading. instead, this content should be included before the *Posts* section.

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -4,10 +4,10 @@ layout: default
 
 <div class="home">
 
-  <h1 class="page-heading">Posts</h1>
-  
   {{ content }}
 
+  <h1 class="page-heading">Posts</h1>
+  
   <ul class="post-list">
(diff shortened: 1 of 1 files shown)


benbalter:
@dret can you provide a screenshot of the before/after here? I'm having trouble following what's going on. Thanks.

dret:
On 2017-04-08 02:37, Ben Balter wrote:
> @dret <https://github.com/dret> can you provide a screenshot of the
> before/after here? I'm having trouble following what's going on. Thanks.

http://webconcepts.info/ is an example of a page that uses a familiar 
flow of having some content pulled in from the page, and then listing 
recent posts using generated content. this seems much more likely to be 
required than having a page that at the very top lists recent posts, and 
only after that lists the actual content of the page.

benbalter:
@jekyllbot: merge +minor.

PR #90, opened 2016-12-05
Add support for excerpts on homepage

ashmaroli:
### Add support for displaying excerpts on the Home Page.
- add `show_excerpts = enabled`  to your _config.yml to enable displaying excerpts on the home page.
- Enabling excerpts adds an underline to the heading and adds an invisible `<hr>` (horizontal rule) to provide a visual separation between list items.

#### Preview of Home Page with excerpts enabled:

excerpts


README.md
```diff
@@ -75,6 +75,16 @@ To enable Google Anaytics, add the following lines to your Jekyll site:
...
 Google Analytics will only appear in production, i.e., `JEKYLL_ENV=production`
 
+--
+
+### Enabling Excerpts on the Home Page
+
+To display post-excerpts on the Home Page, simply add the following to your `_config.yml`:
+
+```yaml
+show_excerpts: enabled
```
benbalter: Should excerpts just be enabled in all cases? Enabled by default? Why do we need to make this an option?

ashmaroli: I'd prefer excerpts to be enabled by default. But made it an option because I didn't want Minima to suddenly change how existing sites would look.. We can make it enabled by default in the next major release.

benbalter: Thanks makes sense.

benbalter: Should this be `true/false` if it's a boolean value?

ashmaroli: `true/false` seems more appropriate :+1:

ashmaroli:
@benbalter  would it be wiser to move the switch to under `minima` namespace? as in
```
{% if site.minima.show_excerpts %}
[...]
```

benbalter:
> would it be wiser to move the switch to under minima namespace?

Not strongly opinionated. If we don't namespace it, we can create a default practice/pattern for other themes.

ashmaroli:
> If we don't namespace it, we can create a default practice/pattern for other themes.

I agree. :+1: 
Ready to be reviewed..

ashmaroli:
#### Revised preview

minima-excerpts

_layouts/home.html
```diff
@@ -16,18 +16,17 @@ <h2 class="post-list-heading">{{ page.list_title | default: "Posts" }}</h2>
...
+          </h3>
+          {% if site.show_excerpts %}
+            {{ post.excerpt }}
+          {% endif %}
+        </li>
+      {% endfor %}
   </ul>
-
-    <p class="rss-subscribe">subscribe <a href="{{ "/feed.xml" | relative_url }}">via RSS</a></p>
+
```
ashmaroli: please address the alignment and extra whitespace on this template before you cut a new version..

ziutek:
I miss something like "More" or  "Continue reading →" link at and of excerpt. If excerpt is long, as you can see for example on https://hackaday.com/blog/ , someone can thing that this is whole small article and don't read full one.

PR #97, opened 2017-01-03
Add disqus on example config

mvcds:
I jumped straight to the example folder's `_config.yml` so I could check how to configure disqus information, as it is the default comment system but I only could found how to configure it on the README.

I think it would be nice to set it on the example and it does not hurt me to open a PR with that, even if the commiters disagree.

### DIFF
--- a/example/_config.yml
+++ b/example/_config.yml
@@ -8,7 +8,8 @@ description: > # this means to ignore newlines until "baseurl:"
 baseurl: "/minima"
 twitter_username: jekyllrb
 github_username:  jekyll
-disqus: # if you want to use disqus as comment system
+# if you want to use disqus as comment system, be aware that it only works on production
+disqus:
   shortname: your_disqus_shortname
 
 # Build settings


example/_config.yml
```diff
@@ -8,7 +8,7 @@ description: > # this means to ignore newlines until "baseurl:"
 baseurl: "/minima"
 twitter_username: jekyllrb
 github_username:  jekyll
-disqus:
+disqus: #if you want to use disqus as comment system
```
ashmaroli: always leave a space after the hash (`#`) in comments.

ashmaroli:
Users tend to fork the repo and serve the 'example ' site to preview the theme.  
`disqus` was *probably* never included in the 'example' site to avoid confusion as the functionality works only in `production`.

mvcds:
What about telling disqus only works on prod?

example/_config.yml
```diff
@@ -8,6 +8,9 @@ description: > # this means to ignore newlines until "baseurl:"
 baseurl: "/minima"
 twitter_username: jekyllrb
 github_username:  jekyll
+# if you want to use disqus as comment system, be aware that it only works on production
+disqus:
+  shortname: your_disqus_shortname
```
benbalter: Should this be commented out by default, such that we don't have a bunch of sites, which by default, have a broken comment system? Also, perhaps we should provide details on how to sign up?

ashmaroli: I still feel there's no place for `production`-only features to be included in the `example/` as long as the example is going to be served only locally. though, I hv never tried disqus with `//localhost`

benbalter: Couldn't we limit the output to production via liquid, and make it opt-in entirely by commeting out the config (and making it dependent on the config existing)?

ashmaroli:
we no longer have an `example/`.. do we??

ashmaroli:
@benbalter this PR is no longer valid since we do not have an `example` directory

example/_config.yml
ashmaroli: > Should this be commented out by default, such that we don't have a bunch of sites, which by default, have a broken comment system?

Yes, lets comment this out by default. Also include a link to disqus registration page.

> Couldn't we limit the output to production via liquid..

That's how [it is](https://github.com/jekyll/minima/blob/master/_includes/disqus_comments.html#L1) right now. And its also dependent on page-wise configuration

DirtyF:
Closing this as we it's properly documented in the README.

PR #99, opened 2017-01-12
Change discuss script URL pattern

skalinets:
For some reason when URL for Disqus script starts from 'https://' comments are
not shown. I changed it to start from '//' like [instructions](https://disqus.com/admin/install/platforms/universalcode)
say and now it works fine. Not sure what was the problem, but maybe this change
will help someone else.

### DIFF
--- a/_includes/disqus_comments.html
+++ b/_includes/disqus_comments.html
@@ -10,7 +10,7 @@
     (function() {
       var d = document, s = d.createElement('script');
 
-      s.src = 'https://{{ site.disqus.shortname }}.disqus.com/embed.js';
+      s.src = '//{{ site.disqus.shortname }}.disqus.com/embed.js';
 
       s.setAttribute('data-timestamp', +new Date());
       (d.head || d.body).appendChild(s);


pathawks:
How are you hosting your site? Is it served over HTTPS?

Using protocol-relative URLs [is an anti-pattern](https://www.google.com/amp/s/jeremywagner.me/blog/amp/stop-using-the-protocol-relative-url?client=safari), so it’d be nice if a better solution could be worked out.

skalinets:
I am hosting my site on [github](http://skalinets.github.io/install-elk-windows.html). I tried both https and http and neither worked. 
However after I reverted changes back to "//", it has stopped working at all. Now neither original minima's template nor mine "patched" version don't work. Despite the fact that latter one was working before. 

Also I noticed that *absolute_url* filter does not work as expected on github: 
```
 var disqus_config = function () {
      this.page.url = '/install-elk-windows.html';
      this.page.identifier = '/install-elk-windows.html';
    };
```

Seems like this is causing issues. Anyway I am canceling this PR since it doesn't seem to fix the issue.

PR #78, opened 2016-11-09
Scale headings relative to base-font-size

lunaryorn:

PR #103, opened 2017-01-25
Use relative size for fonts and line-height

### LINKED ISSUE #102: Blockquote hard-codes font size
The blockquote tag hard-codes the font-size to 18px.  Shouldn't it scale relative to $base-font-size instead, e.g. `$base-font-size * 1.125`?

DirtyF:
Fix #102


ashmaroli:
@DirtyF  I think it would be better if we create a `mixin` for relative font sizes.. and related parameters ..
whatsay?

DirtyF:
@ashmaroli I implemented your suggestion, we could also abstract values and use a typographic modular scale and made it easier for users to change typography for instance, but I don't feel like it's the intent of this theme.

ashmaroli:
I think you misunderstood my intent.
The reason I wanted this to be handled by a mixin was to avoid having to calculate and supply ratios...

I was thinking more along the lines of:
e.g.:

```sass
$base-font-size: 16; /* no units */

@mixin font-size($size) {
  font-size: 0px + $size; /* fallback for older-browsers */
  font-size: 0rem + $size / $base-font-size;
}
```
```sass
body {
  font-size: 0px + $base-font-size;
}

/* use mixin on all other elements. */
h1 {
  @include font-size(32) /* just the size in px. no units */
}
```
will yield:
```css
h1 {
  font-size: 32px; /* fallback for older-browsers */
  font-size: 2rem;
}

DirtyF:
yes I saw that [codepen tutorial](https://css-tricks.com/snippets/css/less-mixin-for-rem-font-sizing/), we can go with [`rem`](http://caniuse.com/#feat=rem) unit, but my intent here was to gently move to a [modular scale](https://github.com/modularscale/modularscale-sass).

ashmaroli:
I was not aware of the modular scale even though it seems to have been around for years. Thank you for the links.
While I feel having to come up with the appropriate ratio is an added complexity for the user planning to customize the stylesheets, I'll leave it to your judgement.

benbalter:
@DirtyF which approach is better? This approach or the approach taken in https://github.com/jekyll/minima/pull/78? Not strongly opinionated. If you think this is the better approach, please resolve the merge conflicts.

DirtyF:
@benbalter #78 was a good start, this PR applies the same principle to other font-size declarations. The mixin is just the way to keep it DRY.

benbalter:
@jekyllbot: merge +minor.

Thanks @DirtyF!

PR #108, opened 2017-02-15
Replace `twitter_username` with `twitter:username`

caspervonb:
The `jekyll-seo-tag` plugin already defines a format where it expects to
find the twitter username.

This replaces all occurences of `twitter_username` with
`twitter:username` to match that expectation.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -28,9 +28,9 @@
           </li>
           {% endif %}
 
-          {% if site.twitter_username %}
+          {% if site.twitter.username %}
           <li>
-            {% include icon-twitter.html username=site.twitter_username %}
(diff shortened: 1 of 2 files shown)


ashmaroli:
This looks like a breaking change for those not using the seo-tag plugin. 

Plus, you may want to create a similar PR for upstream Jekyll Core to have the change reflect in a site generated by `jekyll new`

benbalter:
I'm in favor of this change (it matches the behavior of Jekyll SEO tag), but we should make it a non-breaking changing by falling back to the old key if the new key doesn't exist.

caspervonb:
I'll try to get around to that sometime during this coming weekend :+1:

PR #116, opened 2017-04-06
Fixes #115 - Only include the nav tag if there are pages to be displayed in the menu

### LINKED ISSUE #115: The hamburger is displayed even if there are no site.pages
If there are no site.pages defined, the hamburger menu serves no purpose. Clicking it will do nothing.
Suggesting not including the nav tag at all if there are no links to populate it with.

Adyg:
This checks if there are any site.pages and only displays the nav tag if there are any.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -3,24 +3,25 @@
   <div class="wrapper">
 
     <a class="site-title" href="{{ "/" | relative_url}}">{{ site.title | escape }}</a>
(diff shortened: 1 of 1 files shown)


benbalter:
Thanks @Adyg!

@jekyllbot: merge +minor.

alxn:
@Adyg I wonder if it should only display the hamburger if `>1` item?

PR #117, opened 2017-04-07
More easily customizable CSS

benbalter:
This is an alternate approach to https://github.com/jekyll/minima/pull/86 which mirrors the approach we take on all the https://github.com/pages-themes themes.

The idea being, the default CSS file includes only an @import statement with the themes name, and all defaults and imports exist in the `_sass` directory.

That way a novice user can simply copy the `assets/style.scss` file (or write the one line themself) and add any custom css after the import statement.

The variable defaults already exist in `_sass/minima.scss` and are duplicative (and confusing/overly complex IMHO).

### DIFF
--- a/_sass/minima.scss
+++ b/_sass/minima.scss
@@ -1,3 +1,5 @@
+@charset "utf-8";
+
 // Define defaults for each variable.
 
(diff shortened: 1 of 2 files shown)


benbalter:
@jekyllbot: merge +minor.

benbalter:
Thanks @benbalter!

PR #118, opened 2017-04-07
Remove the Rakefile

benbalter:
Now that #76 has merged, thanks to @DirtyF, we can remove the Gemfile and custom tooling cruft and rely on vanilla-Jekyll to build and serve.


benbalter:
@jekyllbot: merge +dev.

PR #120, opened 2017-04-07
Update default local URL

### LINKED ISSUE #119: localhost:4000/minima does not resolve correctly per README
The readme says:

>To test your theme, run bundle exec rake preview and open your browser at ` http://localhost:4000/minima/`.

...however that does not resolve correctly:

image

...but `http://localhost:4000` does resolve to the theme preview instead:

image

Not sure if this is something specific to my machine, or if we should update the README?

DirtyF:
fix #119

/cc @jglovier


benbalter:
Nice. Thanks @dirtyf!

@jekyllbot: merge +docs.

PR #121, opened 2017-04-08
Add 404 page from jekyll `master`

ashmaroli:
Now that Minima is live on Github Pages and Jekyll itself will provide this template from `v3.5`, let's display the custom `404.html` instead of using the one provided by Github

### DIFF
--- a/404.html
+++ b/404.html
@@ -1,0 +1,24 @@
+---
+layout: default
+---
+
+<style type="text/css" media="screen">
+  .container {
+    margin: 10px auto;
+    max-width: 600px;
+    text-align: center;
+  }
+  h1 {
+    margin: 30px 0;
+    font-size: 4em;
+    line-height: 1;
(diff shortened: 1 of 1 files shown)


benbalter:
@jekyllbot: merge +minor.

Thanks @ashmaroli!

adamvoss:
When using the minima gem with GitHub Pages, am I correct this file needs to be explicitly duplicated in the site's repo?  That is, the version in the gem will not be served?

ashmaroli:
> am I correct this file needs to be explicitly duplicated in the site's repo? That is, the version in the gem will not be served?

@adamvoss Yes. This file is not part of the theme-gem. I intended it to demo the new `404.html` page that'll be shipped with **`Jekyll v3.5`**. Sorry for not mentioning it explicitly. Once J-3.5 is out, all new projects will have this page included automatically. For existing projects, its left to the owner's discretion.

PR #122, opened 2017-04-08
Bigger font-size for h1 in content

### LINKED ISSUE #113: h2 has a larger font than h1 in Chrome
I see in layout.scss that Minima is forcing `.post-content h2` to 32px. However, I don't see a similar rule for `.post-content h1`. This means that by default, in Chrome, `h2` elements are larger than `h1`s, because Chrome's default size for `h1` is 24px. Is this intentional?

DirtyF:
fix #113


ashmaroli:
its not just for posts. the `page` layout uses `.post-content` as well

DirtyF:
@ashmaroli I don't know if it's a good idea to encourage many h1 per page (I am already sceptical for posts)

ashmaroli:
> I don't know if it's a good idea to encourage many h1 per page

Its not. The idea even is frowned upon by the community. But regular users need not know about such conventions.

Btw, I was hinting at this PR's Title. This PR actually applies to both `post` layout and `page` layout as they both reference a `<div class="post-content">`

antgel:
> The idea even is frowned upon by the community. But regular users need not know about such conventions.

Where does that leave this PR? :)

dvhart:
As a new user, I just ran into this myself today and eventually made my way here. Nothing in my reading of the jekyll docs about writing posts or in the markdown syntax hinted that there should not be more than # or === header per document. If there is some thing that documents this as "bad" other than tribal knowledge, I missed it. This seems like a reasonable fix to push to help "regular users" like myself avoid this issue.

dvhart:
Perhaps the issue is with the mapping of a MD H1 header # or === to an HTML <h1> tag? Since we don't use an MD header marker for the title in the pages themselves (that's in the filename or the front matter as I understand it), then perhaps Jekyll should be mapping MD HN header tags to <hN+1> in html. (MD H1 becomes <h2> in HTML). This would effectively reserve <h1> for the title by mapping MD tags to <h2> and lower. Perhaps it's too late for that given the number users now dependent on the existing behavior.

ashmaroli:
@DirtyF So many merge commits!! :roll_eyes: 
I'd just squash all those superfluous commits into one single merge commit..

DirtyF:
fixed in #273

PR #124, opened 2017-04-10
added support for txtpen

rickyhan:


caspervonb:
Seems unlikely that this will go in, but more commentary than "adds support for my thing" would probably be appreciated here.

Seems to be a service, which adds excerpt highlighting similar to medium selections?

rickyhan:
txtpen is the only service that provides highlighting and in-text commenting similar to Medium. I think adding txtpen in the official documentation will benefit new Jekyll users and bring web annotation to more people.

Thanks in advance.
Ricky

benbalter:
@rickyhan Thanks for the Pull Request! What's unique about txtpen that would make it a good addition to the default site for all Jekyll users? "This will benefit users" is not the same as "this should be part of the default". If we used that line of reasoning for every web service the theme would quickly become unmaintainable. A quick GitHub search suggests [only 8 sites on GitHub](https://github.com/search?l=HTML&q=txtpen.com&ref=simplesearch&type=Code&utf8=%E2%9C%93) use it today.

PR #128, opened 2017-05-08
Provide exclude mechanism for minima

dirkgomez:
Put this into your front matter to exclude pages:

exclude: true

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -21,7 +21,7 @@
         <div class="trigger">
           {% for path in page_paths %}
             {% assign my_page = site.pages | where: "path", path | first %}
-            {% if my_page.title %}
+            {% if my_page.title and my_page.exclude_from_nav != true %}
             <a class="page-link" href="{{ my_page.url | relative_url }}">{{ my_page.title | escape }}</a>
             {% endif %}
           {% endfor %}
--- a/README.md
+++ b/README.md
@@ -132,6 +132,17 @@ To enable Google Anaytics, add the following lines to your Jekyll site:
 
 Google Analytics will only appear in production, i.e., `JEKYLL_ENV=production`
 
+--
+
+### Excluding pages from the default navigation
+
+To exclude a page from the default navigation, put this into that page's front
+matter:
+
+```yaml
+  exclude_from_nav: true
+```
+
 ## Contributing
 
 Bug reports and pull requests are welcome on GitHub at https://github.com/jekyll/minima. This project is intended to be a safe, welcoming space for collaboration, and contributors are expected to adhere to the [Contributor Covenant](http://contributor-covenant.org) code of conduct.


ashmaroli:
To exclude a page from being processed, simply add `published: false` to the front matter

pathawks:
@ashmaroli It looks like this is about excluding a page from the default navigation, rather than preventing it from being published.

_includes/header.html
```diff
@@ -21,7 +21,7 @@
         <div class="trigger">
           {% for path in page_paths %}
             {% assign my_page = site.pages | where: "path", path | first %}
-            {% if my_page.title %}
+            {% if my_page.title and !my_page.exclude != true %}
```
benbalter: I'd recommend this be something more descriptive, like `exclude_from_nav` to prevent collisions with existing YAML.

ashmaroli:
> It looks like this is about excluding a page from the default navigation

@pathawks You're right. My bad..

_includes/header.html
dirkgomez: Where in the docs would I put that? I can't find a section that seems a good fit to me (and I'm a Jekyll newbie, so a bit lost anyway).

Strangehill:
I find whitelisting pages with an include field rather than blacklisting with an exclude more useful more often.

dirkgomez:
Wouldn't that be a breaking change though?

Strangehill:
Oh, that's right. Good point.

_includes/header.html
adamvoss: @dirkgomez Looking at the existing content, I would expect to see it as a new **h3** in the [Usage](https://github.com/jekyll/minima#usage) section

_includes/header.html
```diff
@@ -21,7 +21,7 @@
         <div class="trigger">
           {% for path in page_paths %}
             {% assign my_page = site.pages | where: "path", path | first %}
-            {% if my_page.title %}
+            {% if my_page.title and !my_page.exclude_from_nav != true %}
```
benbalter: Shouldn't this be `== true`?

ashawley: I think the issue is the unary not (`!`), which I understand is not supported in liquid.  I believe this is sufficient:

```
if my_page.title and my_page.exclude_from_nav != true
```

Unfortunately, `== false` isn't available to us here because of a defect with truthiness in liquid/ruby with respect to booleans, nulls, empty string, uninitialized values, and so on...

adamvoss: @dirkgomez I tested and confirmed @ashawley's logic to be correct.  In the current state (8fb7b81) the boolean is working opposite from desired behavior.

Also, it is up to you if you care or not, but I noticed your first commit (41529ce2f614a540c74d1af1d216a9cfa79cea58) was done with a different e-mail address than the others, and that e-mail address is not associated with your GitHub account.  You may wish to squash/rebase (then force-push) to bring them all under one address or to add your other e-mail address to your GitHub account so it is recognized as yours.

_includes/header.html
```diff
@@ -21,7 +21,7 @@
         <div class="trigger">
           {% for path in page_paths %}
             {% assign my_page = site.pages | where: "path", path | first %}
-            {% if my_page.title %}
+            {% if my_page.title and my_page.exclude_from_nav == true %}
```
adamvoss: Should be,
```ruby 
and my_page.exclude_from_nav != true
```
Thanks for sticking with this.  I know from experience getting boolean logic right can be a pain :-)

dirkgomez:
Amended the commits accordings - and thank you for the patience, still grappling with Jekyll.

ashawley:
Why isn't whitelisting with `header_pages` sufficient?

dirkgomez:
Looks very similar indeed, how do I use header_pages?

ashawley:
Looks like #52 documented it in the config but not the README

dirkgomez:
ashawley is right, I'm closing this pull request.

ashawley:
For the record, I wasn't opposed to this change.  I was just wondering why the question wasn't raised about the overlap with `header_pages`.   It takes a lot of humility to close a PR and then even further offer to make the documentation fix.  Good on you, dirkgomez!

To be honest, I would still entertain this improvement.  I predict it will come up again for scenarios where blacklisting may be easier to maintain than whitelisting (with `header_pages`).

adamvoss:
#134 continues from the discussion here by adding documentation for `header_pages` to **README.md.
(Commenting so the issues are linked on GitHub.)

@ashawley Like @dirkgomez I had need for something like this.  I agree in not being opposed to this, but I am not sure there is enough reason to push for it now either.  If you look at the behavior of minima when there are lots of nav links, I think from a practical standpoint the number of desired nav links will be kept low, favoring a whitelist approach.  Plus, `header_pages` gives the added benefit of defining order.

The main advantage of the approach in this PR would be the resilience to file moves/renames.

Issue #98, opened 2017-01-04
Broken link to /feed.xml out-of-the-box.


gsnedders:
Out of the box, minima produces a broken link to `/feed.xml`, because it requires jekyll-feed. Minima should either depend on jekyll-feed or not output the link to the feed.

pathawks:
For reference, [here is the offending line](https://github.com/jekyll/minima/blob/v2.1.0/_includes/head.html#L11)

DirtyF:
@pathawks I was thinking of adding jekyll-feed as a runtime dependency in the gemspec, modify example Gemfile and _config.yml and use feed meta instead in the include, do this seem OK to you?

On a side note, an `html-proofer` test of the output wouldn't hurt to avoid this.

pathawks:
I seem to remember their being an intentional decision _not_ to tie **minima** to **jekyll-feed**, as **minima** is supposed to be a minimal.

Can't find that thread now.

gsnedders:
I for one would rather it not be a dependency, FWIW, especially for sites which don't have any posts (i.e., consist merely of pages and collections).

alzeih:
As a workaround, I used this in `_layouts/home.html`:

```
  {% if site.gems contains "jekyll-feed" %}
    <p class="rss-subscribe">subscribe <a href="{{ "/feed.xml" | relative_url }}">via RSS</a></p>
  {% endif %}
```

alzeih:
There's (only) one more place that references feed.xml:

`_includes/head.html`contains

```  <link rel="alternate" type="application/rss+xml" title="{{ site.title | escape }}" href="{{ "/feed.xml" | relative_url }}">```

So perhaps this should be guarded with a check also?

ashawley:
If you use `jekyll new`, then you get the dependency for `jekyll-feed`:

See https://github.com/jekyll/jekyll/blob/af5d096/lib/site_template/_config.yml#L31

So someone just forgot to add that dependency when the theme was forked to this repo last year?

ashmaroli:
> @DirtyF: I was thinking of adding jekyll-feed as a runtime dependency
> @pathawks: ..an intentional decision not to tie **minima** to **jekyll-feed** as **minima** is supposed to be a minimal.

Is the above still valid, esp. since Minima will now come with `jekyll-seo-tag` included?

alzeih:
I think minima should really be minimal, and not require jekyll-feed or jekyll-seo-tag (or any other jekyll plugins).

[...]

PR #129, opened 2017-05-12
Change RSS references

Cj-Malone:
The jekyll-feed plugin uses Atom not RSS.

But I've removed the format completely as I don't think people care. I also fixed the mime type, which should help with some parsing issues.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -8,7 +8,9 @@
 
   <link rel="stylesheet" href="{{ "/assets/main.css" | relative_url }}">
   <link rel="canonical" href="{{ page.url | replace:'index.html','' | absolute_url }}">
-  <link rel="alternate" type="application/atom+xml" title="{{ site.title | escape }}" href="{{ "/feed.xml" | relative_url }}">
+  {% if site.gems contains "jekyll-feed" %}
+    <link rel="alternate" type="application/atom+xml" title="{{ site.title | escape }}" href="{{ "/feed.xml" | relative_url }}">
+  {% endif %}
   
   {% if jekyll.environment == 'production' and site.google_analytics %}
   {% include google-analytics.html %}
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -8,7 +8,7 @@
 
   <link rel="stylesheet" href="{{ "/assets/main.css" | relative_url }}">
   <link rel="canonical" href="{{ page.url | replace:'index.html','' | absolute_url }}">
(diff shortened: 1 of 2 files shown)


pathawks:
This should probably just be removed, as Minima does not generate _any_ feed out of the box.

Cj-Malone:
@pathawks I've just seen https://github.com/jekyll/minima/issues/98 too, I've added the checks that @alzeih recommended. I think it's best to have the feed references added by default in minima for users that opt to use `jekyll-feed`. It's best for it to be done right here rather than an implementation per site.

aterenin:
Might I suggest using a SVG icon instead of the word `Subscribe`? I think it would look a bit cleaner and more in line with the overall theme (similar to GitHub, Twitter icons). See `http://avt.im` for an example. I am happy to make a PR if people agree.

ashmaroli:
> using a SVG icon instead of the word Subscribe?

That would appear less like a link. A bit too minimalistic?
On the other hand, some minimal text with the icon would look fine IMO..

alzeih:
@Cj-Malone thanks! Jekyll seems to have renamed `site.gems` to `site.plugins` (https://github.com/jekyll/jekyll/issues/4509 , https://github.com/jekyll/jekyll/pull/5130 , https://github.com/jekyll/jekyll/pull/6045 ) so maybe replace `site.gems` with `site.plugins` ?

As Jekyll plugins can alternatively be enabled in Bundler (https://jekyllrb.com/docs/plugins/), the plugin may be enabled and not be listed in `site.plugins`. I haven't found a way to query gems in group `:jekyll_plugins` from liquid yet. Does someone know a way?

ashmaroli:
> Jekyll seems to have renamed `site.gems` to `site.plugins`

Yes. But its only been deprecated in Jekyll. There still may be sites out there using `Jekyll 3.5.x` but with `gems:` in their config file..
So its bit of a fragmentation issue now.

But if Team Minima decides to include this feature as part of Minima 3.0, then its probably safe to disregard Jekyll-3.5.x users still on the `gems:` config

DirtyF:
> As Jekyll plugins can alternatively be enabled in Bundler (https://jekyllrb.com/docs/plugins/), the plugin may be enabled and not be listed in site.plugins.

True. Closing them as this workaround does not address the entire issue.

ashawley:
Ok, I didn't realize it was impossible to sense plugins in Jekyll in all situations.  Seems that the safest way to fix #98 is to break the feed by deleting it.

PR #130, opened 2017-05-22
refactor scss to mobile-first

Strangehill:
media-queries are 'mobile-second' and this pr tries to keep all the same styles but turning media queries inside out.


benbalter:
@Strangehill thanks for this. Is there any visible change? if so, can you provide screenshots?

Strangehill:
I tried to avoid any visible changes.

_sass/minima.scss
```diff
@@ -21,18 +21,18 @@ $grey-color-dark:  darken($grey-color, 25%) !default;
 // Width of the content area
 $content-width:    800px !default;
 
-$on-palm:          600px !default;
-$on-laptop:        800px !default;
+$on-medium:        600px !default;
+$on-big:           800px !default;
```
ashmaroli: I would prefer seeing `$on-large`.. whatsay?

_sass/minima/_layout.scss
```diff
@@ -25,77 +25,100 @@
...
-      display: block;
+      display: inline;
       float: right;
-      width: 36px;
-      height: 36px;
+      width: auto;
+      height: auto;
       z-index: 2;
       cursor: pointer;
     }
```
ashmaroli: you have declared `display: none` for `.menu-icon ` @ `$on-medium` in the following lines.. so this `label [for="nav-trigger"]` is essentially empty on `$on-medium` (and beyond), isn't it?

_sass/minima/_layout.scss
```diff
@@ -25,77 +25,100 @@
...
-      clear: both;
       display: none;
     }
 
-    input:checked ~ .trigger {
+
+    //label[for="nav-trigger"] {
+     // display: inline;
+    //}
+
```
ashmaroli: this comment could be removed..

_sass/minima/_layout.scss
```diff
@@ -129,44 +152,44 @@
 }
 
 .footer-col {
-  float: left;
+  float: none;
```
ashmaroli: this `float: none` is being overridden media queries at `$on-medium` and beyond.
on `$on-small` this would bring the social-icons column vertically between the author and description, If that's the intention, leaving out the `float` is better than `float: none`

_sass/minima.scss
Strangehill: sounds great!

_sass/minima/_layout.scss
Strangehill: oops, will get rid of that

_sass/minima/_layout.scss
Strangehill: Ah, yes. And so, I'm removing all styles for `label [for="nav-trigger"]` , except for `display` which should be `display:none` like `.menu-icon` @ `$on-medium`

_sass/minima/_layout.scss
Strangehill: Agreed, that was superfluous.

_sass/minima.scss
```diff
@@ -21,18 +21,18 @@ $grey-color-dark:  darken($grey-color, 25%) !default;
 // Width of the content area
 $content-width:    800px !default;
 
-$on-palm:          600px !default;
-$on-laptop:        800px !default;
+$on-medium:        600px !default;
+$on-large:           800px !default;
```
ashmaroli: extra whitespace here..

Strangehill: got it

ashmaroli:
Milestone: `3.0.0`

ashmaroli:
@DirtyF This is a breaking change. I recommend this be added to a `3.0` milestone

DirtyF:
@ashmaroli I don't consider this a breaking change, it just inverts the logic in the media queries, there's no visual difference on the default content, but fine to wait before 3.0 to go mobile-first.

I just re-tested this to post some screenshots of this version as @benbalter asked:

mobile-first

mobile-first-medium

@Strangehill We are much interested in this and would like to go mobile first in the next major version of minima.

ashmaroli:
The reason I feel this is a breaking-change is because, say, a user has a custom `assets/main.scss` like so:
```sass
@import "minima";
@import "custom";
```
```sass
// _sass/custom.scss

.nav {
  @include media-query($on-palm) {
     position: absolute;
     top: 9px;
     right: $spacing-unit / 2;
     background-color: $background-color;
  }
}
```

In theory, this *could* break the site because, Minima no longer defines the variable `$on-palm`

ashawley:
> In theory, this _could_ break the site because, Minima no longer defines the variable `$on-palm`

So could `$on-palm` be preserved?  Or does it need to be deleted?

Strangehill:
I could put `$on-palm` back in (alongside `$on-medium` and `$on-large`). Could we then get the merge sooner rather than later and remove `$on-palm` for 3.0.  
The same could be done for `$on-laptop`.

ashmaroli:
> So could `$on-palm` be preserved?

Yes it can be preserved.. and so can, the other similar variables.
The question here is, do we need to.., if this is going to be included in `v3.0`..?

ashmaroli:
> Could we then get the merge sooner rather than later and remove $on-palm for 3.0.

/cc @DirtyF @benbalter

ashawley:
Yeah, I'd suggest getting it merged sooner in the 2.x series

Strangehill:
Also worth noting: the media query mixin seems as much of a breaking change as the two breaking point variables since this mixin too could have been put to use by users. 
```
@mixin media-query($device) {
-  @media screen and (max-width: $device) {
+  @media screen and (min-width: $device) {
     @content;
   }
 }
```

DirtyF:
We build for the 90% here, minima is a default theme not a starter theme. 
It there is any breaking change it should be documented in an upgrade section of the docs.

ashmaroli:
@Strangehill  why not insert a new mixin for the media-queries like the existing one..?

Strangehill:
I've restored the media query mixin that existed, but because it is desktop-first, I have also removed the instances where it's used. I also restored $on-palm, and $on-laptop, though, like the mixin, they're not in use anywhere and should be removed only until 3.0, to avoid breaking changes before then.
@ashmaroli Not sure what the original mixin was bringing to the table in the first place. Would you recommend inserting a new one?

ashmaroli:
> Would you recommend inserting a new one?

With the mixin, one needn't bother about adding `@media screen and (max-width: $device)` everywhere, they could simply do an `@include { }` and insert their styles for that breakpoint.

For the present case, I'm thinking something like
```sass
// Use mobile-first media queries like this:
// @include mobile-media-query($on-palm) {
//   .wrapper {
//     padding-right: $spacing-unit / 2;
//     padding-left: $spacing-unit / 2;
//   }
// }
@mixin mobile-media-query($device) {
  @media screen and (min-width: $device) {
    @content;
  }
}
```
pretty much the same thing as before, just a new one till the Team ships a `3.0` when we'll remove the older mixin and rename the newer one.. .

_sass/minima.scss
```diff
@@ -23,6 +23,9 @@ $table-text-align: left !default;
 // Width of the content area
 $content-width:    800px !default;
 
+$on-medium:        600px !default;
+$on-large:         800px !default;
```
ashmaroli: instead of adding these two declarations here, simply do the following after existing variables:
```scss
$on-medium:    $on-palm !default;
$on-large:     $on-laptop !default;
```

Strangehill:
re: a new mixin, Going from:
`@media screen and (max-width: $device){ }` to 
`@include mobile-media-query($device){ }` would not seem to require a mixin though. What is earned other than the typing of a handful of characters?

ashmaroli:
> What is earned other than the typing of a handful of characters?

The respite from confusion due to overlooking the fact that the newer explicit `media-queries` are set to `min-width` instead of `max-width`.

But then again, clarity in documentation would be better too.. I defer this to the judgment of the maintainers here..

Strangehill:
that's definitely a possibility. Would it be acceptable to have a comment right above the old mixin?
e.g.
``` 
// Notice the following mixin uses max-width, in a deprecated, desktop-first
// approach, whereas media queries used elsewhere now use min-width.
```
This would also be removed for 3.0, at which point I'd suggest the documentation is updated.

ashmaroli:
I like the comment too.. whatsay @DirtyF?

DirtyF:
In the news: https://searchengineland.com/googles-mobile-first-index-rolled-site-will-roll-slowly-285590

PR #132, opened 2017-05-29
HTML escape author in posts

ashawley:


benbalter:
@ashawley thanks for this. Can you provide some more context on what you're trying to do here?

ashawley:
Someone had two author's names on a post, and I noticed that the ampersand wasn't escaped.

DirtyF:
Should @jekyll/minima support multi-author posts and allow to maintain a list of authors?

ashawley:
> support multi-author posts and allow to maintain a list of authors?

That'd be cool.  Wouldn't that need to be agreed upon and implemented in Jekyll, generally?

DirtyF:
@ashawley I was just thinking of specifying authors in `_config.yml` or in a _`data/authors.yml` and make use of this in the template.

PR #134, opened 2017-06-07
Add header_pages to README.md (see issue #52)

dirkgomez:


README.md
```diff
@@ -132,6 +132,17 @@ To enable Google Anaytics, add the following lines to your Jekyll site:
 
 Google Analytics will only appear in production, i.e., `JEKYLL_ENV=production`
 
+### Link to specific files in header
```
adamvoss: Suggested title for consideration: `Customizing navigation links`

README.md
```diff
@@ -132,6 +132,17 @@ To enable Google Anaytics, add the following lines to your Jekyll site:
...
+
+Add this to only link to the about and the linklists page in the default
+navigation header:
+
+```yaml
+header_pages:
+  - about.md
+  - linklists.md
+```
+
```
adamvoss: Suggested addition:

This allows you to set which pages you want to appear in the navigation area and configure order of the links.

README.md
DirtyF: or something like `Defining the main navigation links`

README.md
DirtyF: I agree with @adamvoss, it's best to share the reason why one may want to use this option.

README.md
ashawley: Seems like there was a consensus on **Navigation links**  ;-)

Keep it minima!™

PR #137, opened 2017-06-29
Increase flexibility of 'home' layout

### LINKED ISSUE #136: Only show "posts" heading if there are posts
If you use minima for a non-blog site, the homepage has a "posts" heading that cannot be removed.

ashmaroli:
Fixes #136 

Allow greater flexibility with the `home` layout

### Notable changes
  - Main heading can now be set via `index.md` with front matter key `title`
  - This heading will now have the default size of `2em` from earlier `1.25em`
  - `Posts` is now an `h2` heading and can be altered with fm-key `list_title`
  - The entire Posts Listing and feed-link is now rendered only if the site has 1 or more `posts`

/cc @benbalter

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Minima has been scaffolded by the `jekyll new-theme` command and therefore has a
 Refers to files within the `_layouts` directory, that define the markup for your theme.
 
   - `default.html` &mdash; The base layout that lays the foundation for subsequent layouts. The derived layouts inject their contents into this file at the line that says ` {{ content }} ` and are linked to this file via [FrontMatter](https://jekyllrb.com/docs/frontmatter/) declaration `layout: default`.
-  - `home.html` &mdash; The layout for your landing-page / home-page / index-page.
+  - `home.html` &mdash; The layout for your landing-page / home-page / index-page. [[More Info.](#home-layout)]
   - `page.html` &mdash; The layout for your documents that contain FrontMatter, but are not posts.
   - `post.html` &mdash; The layout for your posts.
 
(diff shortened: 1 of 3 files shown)


_layouts/home.html
```diff
@@ -3,24 +3,28 @@
 ---
 
 <div class="home">
+  {% if page.title %}
```
DirtyF: What's the purpose of adding such a title here?

ashmaroli: Its a slot to insert an optional `<h1>` to the landing page where heading `!= {{ site.title }}`.

e.g.

--
# Blah Chronicles | (site title) |

## The Spiral Consortium | (h1 page title) |

Lorem ipsum dolor sit amet and all that random stuff.
But before we move on, here's a list of all the posts in this site:

### Posts | (h2 list title) |

  - **Jun 27, 2017**
    **Welcome to the Consortium**
    Welcome cadet! By the end of this post, you'll be familiar about what we do here...
  - **Jun 28, 2017**
    **The Alpha**
    She may not look that menacing, but cross her path wrongly and you shall feel her wrath...

ashmaroli:
@DirtyF Added documentation. Requesting a re-review. Thanks

README.md
```diff
@@ -68,6 +68,25 @@ This directory can include sub-directories to manage assets of similar type, and
...
 
+### Home Layout
+
+`home.html` is a flexible HTML layout for the site's landing-page / home-page / index-page. <br/>
+
+#### Main Heading and Content-injection
+
+From Minima v2.2 onwards, the *home* layout will inject all content from your `index.md` / `index.html` **before** the **`Posts`** heading. This will allow you to include non-posts related content to be published on the landing page under a dedicated heading. *We recommended that you title this section with a Heading2 (`##`)*.
+
+Usually the `site.title` itself would suffice as the implicit 'main-title' for a landing-page. But, if your landing-page would like a heading to be explicitly displayed, then simply define a `title` variable in the document's front matter and it will be rendered with an `<h1>` tag.
```
benbalter: When would you want the index's title to not be the site's title? Wondering if this is abstraction for the sake of abstraction or if the added complexity in a starter theme is worth the benefit. 

It would also cause a conflict on GitHub Pages, which sets titles by default from the first heading.

_layouts/home.html
```diff
@@ -3,24 +3,28 @@
...
 
-  <h1 class="page-heading">Posts</h1>
-  
-  <ul class="post-list">
-    {% for post in site.posts %}
-      <li>
-        {% assign date_format = site.minima.date_format | default: "%b %-d, %Y" %}
-        <span class="post-meta">{{ post.date | date: date_format }}</span>
+  {% if site.posts.size > 0 %}
+  <h2 class="post-list-heading">{{ page.list_title | default: "Posts" }}</h2>
```
benbalter: Similar question here. When would you want this to be something other than Posts since the list is hard coded to the Posts collection (and a use sophisticated enough to introduce their own second collection would likely want to modify the template anyway)?

README.md
ashmaroli: > When would you want the index's title to not be the site's title?

Very rarely. That's why it's optional. The `<h1>` tag won't render unless the user wants it to..
Would provide a fallback in the rare scenario that the user's custom stylesheet has a `.home h1` especially since `Posts` is now a `<h2>`

> It would also cause a conflict on GitHub Pages, which sets titles by default from the first heading.

I did consider this. Wouldn't the plugin not mangle the Index title if `page.title` is defined? I fail to see how a conflict would arise in this case.
The first heading previously  (`<h1>Posts</h1>`) is now after the `{{ content }}` which may or may not contain a heading. So this can be a solution for both situations.

_layouts/home.html
ashmaroli: What if the user wanted the list heading to say `All posts` or Posts in another language? Wouldn't it be better entering a front matter key instead of maintaining an entire `home.html` at `source` just to edit this one line?

Issue #96, opened 2016-12-30
Add Support For Meta Descriptions From Front Matter


mpchadwick:
Currently the meta description [either truncates the page excerpt at 160 characters or defaults to the site description](https://github.com/jekyll/minima/blob/bcabdbd5ba83e0688200607610b83b0d6290fe36/_includes/head.html#L7). It is good SEO practice to hand write meta descriptions, rather than just using the first 160 characters of the post. I propose the meta description first check if there is a "description" set in the front matter and, first and foremost, prioritize that. Happy to submit a PR if the maintainers are on board.

ashmaroli:
Seems like a reasonable request to me..
/cc @jekyll/minima

parkr:
@benbalter What do you think about using jekyll-seo-tag here?

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

PR #139, opened 2017-07-05
Add jekyll-seo-tag dependency

### LINKED ISSUE #138: Add Open Graph tags

DirtyF:
From Jeyll 3.5, gem-based themes can declare plugin dependencies.

Let's improve SEO for all `minima` users.

fix #138 #96

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,9 @@ Contains the `main.scss` that imports sass files from within the `_sass` directo
 
 This directory can include sub-directories to manage assets of similar type, and will be copied over as is, to the final transformed site directory.
 
+### Plugins
+
+Minima comes with [`jekyll-seo-tag`](https://github.com/jekyll/jekyll-seo-tag) plugin preinstalled to make sure your website gets the most useful meta tags. See [usage](https://github.com/jekyll/jekyll-seo-tag#usage) to know how to set it up.
 
 ## Usage
 
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -3,13 +3,12 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
 
-  <title>{% if page.title %}{{ page.title | escape }}{% else %}{{ site.title | escape }}{% endif %}</title>
-  <meta name="description" content="{{ page.excerpt | default: site.description | strip_html | normalize_whitespace | truncate: 160 | escape }}">
+  {% seo %}
 
   <link rel="stylesheet" href="{{ "/assets/main.css" | relative_url }}">
-  <link rel="canonical" href="{{ page.url | replace:'index.html','' | absolute_url }}">
+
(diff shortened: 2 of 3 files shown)


ashawley:
For completeness, list it as a plugin in the config?

ashmaroli:
> For completeness, list it as a plugin in the config?

@ashawley this isn't necessary, Jekyll 3.5 will look into the `.gemspec` and automatically `require` the seo plugin.

ashmaroli:
@DirtyF Can we remove the whitespace *around* the `<link>` tags?
Also, it'd be nice if the `README` mentioned this enhancement and advising users there's *no need* for additionally adding the seo-plugin to their config file / Gemfile

ashawley:
> this isn't necessary, Jekyll 3.5 will look into the .gemspec and automatically require the seo plugin.

I recognize it's not necessary.  Just seems worth documenting in the default config

ashmaroli:
> seems worth documenting in the default config, as well as the README

I'm fine with appending to the README but on the fence about documenting in the config file because IMO it wouldn't matter if users add the seo-plugin to their config file because Ruby _may not_ call `require` on the same plugin twice (currently demonstrated with a plugin added to both `:jekyll_plugins` in Gemfile **and** `plugins` in config file.)

ashawley:
> it wouldn't matter if users add the seo-plugin to their config file because Ruby may not call `require` on the same plugin twice

Are you saying that plugins listed under `plugins` in the config automatically get pulled in by Ruby?  I didn't think it did that.  It only gives an error.  Maybe I'm missing something.

ashmaroli:
> Are you saying that plugins listed under plugins in the config automatically get pulled in by Ruby?

Plugins listed under `plugins:` in the config file get pulled in by Ruby if the plugin-gem is available for loading. Jekyll raises errors under following situations:
  - The plugin is not installed on the local system
  - When there's a `Gemfile`, and the installed plugin is not included in it.
The error in this case is due to Bundler selectively masking the installed gems based on the Gemfile, ergo to Jekyll, your plugin-gem is not installed if its not in the `Gemfile`

GitHub Pages ignores a Gemfile, so it loads plugins based on the config file alone. Once it gets upgraded to Jekyll 3.5, it will additionally load `whitelisted` plugins from a theme's `.gemspec`

ashawley:
> When there's a Gemfile, and the installed plugin is not included in it.

Ah, this is the piece I was missing.  Thanks.

DirtyF:
fix #96

DirtyF:
@benbalter For now It works locally with GitHub Pages gem when pointing to minima `master` branch but not on GitHub Pages, build failled.

benbalter:
@DirtyF if we cut a new version of `minima` and bump it in the GitHub Pages Gem, it should build properly (lest any user could add a runtime dependency to a theme and whitelist any gem for use on GitHub Pages).

ashmaroli:
@DirtyF Since you did not invoke JekyllBot to merge this, you'll have to manually update `History.markdown` to document this enhancement.

PR #112, opened 2017-02-27
Generalize and reorganize social media links #39

domingohui:
Based on #39 
Will be adding svg's for the remaining links soon :)

PR #141, opened 2017-07-12
Refactor social links

ashawley:
This makes the required consolidation of social links to a single file as suggested in #112.

It required removing the use of the templates on the about page.  Hopefully, others aren't depending on it.


ashmaroli:
> It required removing the use of the templates on the about page. Hopefully, others aren't depending on it.

The `about.md` here is only for demonstration on GitHub Pages. It's not included in the release. The namesake release-version has already been [fixed](https://github.com/jekyll/jekyll/commit/90d747238870bc178b533428559d1673dc56563e) released with Jekyll 3.5.0.

ashmaroli:
Tested this locally. No major issues.

Just one personal nit:
`viewBox` is not a valid css property (couldn't find it on caniuse.com)

Also, it'd be great if the `README.md` documents these new config keys.
Thanks for opening this PR :+1:

ashawley:
> `viewBox` is not a valid css property

Ok, I wouldn't know the reason for it so I can' defend it, but the SVG icons still work now after it's removed. 👍 

> Also, it'd be great if the README.md documents these

Ok, I've taken a first pass at documenting them in a new section in the README.

README.md
```diff
@@ -122,6 +121,27 @@ If you don't want to display comments for a particular post you can disable them
...
+```yaml
+twitter_username: jekyllrb
+github_username:  jekyll
+dribbble_username: dribbble
+facebook_username: fb
+flickr_username: flickr
+instagram_username: ig
+linkedin_username: linkedin
+pinterest_username: pinterest
+stackoverflow_username: stackoverflow
```
ashmaroli: since this is documentation, I think we'll have to point to the right way as well.
e.g. StackOverflow profiles are located at `users/` i.e. for `username: minima`, the profile is actually at `stackoverflow.com/users/[random]/minima`

ashawley: Agreed, I'll try to improve them.

I forgot that Stack Overflow was like this.  Unfortunately, the premise of these social links is that the username doubles as both the link and the link text.  :-(

We could of course minimize to this:
```yaml
stackover_username: 111111/jekyll
```
But the link text would still be **111111/jekyll** unless it was filtered out.  Maybe punt on Stackoverflow for now?  See if people actually want it?  Stack Overflow have their own flair badges anyway, see: https://stackoverflow.com/users/flair

ashmaroli: I'm not overly concerned about it. Just mentioned it as an afterthought.. :smiley:

ashawley:
Since the link text is goofed up, I'm going to drop Stackoverflow unless there are any objections...

screen shot 2017-07-19 at 2 28 13 pm

ashmaroli:
@ashawley, what do you think about changing the format entirely?
  - Convert social-links into a **horizontal "social-bar"**
    - last section of the footer
    - full width
  - Drop the names but instead increase the icon-size a bit
  - The names can be added as the `title` attribute for the links/icons.

this is in line with minimalism and also will minimally alter the footer height with addition of multiple icons.

ashawley:
As a way to get around the stackoverflow issue?  I guess that could work.

I may not have the design expertise to make it work.

I've rebased and resolved the latest conflicts.  

Do you want to fork this branch and give it a try in yet another PR?

ashmaroli:
I started a discussion for Minima 3.0 to include breaking changes (such as the idea I proposed above)

> Do you want to fork this branch and give it a try in yet another PR?

That could lead to messy git history..
IMO, you may continue refining this PR as you see fit.

I'll start fresh.

ashawley:
> That could lead to messy git history..

I'm just trying to get people's contributions across the finish line.  I'm happy to clean up git history if that's what keeping this from getting merged.

ashmaroli:
> I'm just trying to get people's contributions across the finish line

You misunderstood me.
This PR is good enough. But I'm starting fresh because I'm introducing structural changes to the footer which may not be approved immediately. Like i said it's targeted  for a future 3.0 version

Hence starting fresh.

about.md
```diff
@@ -7,9 +7,7 @@ permalink: /about/
...
 
 You can find the source code for the Jekyll new theme at:
-{% include icon-github.html username="jekyll" %} /
-[minima](https://github.com/jekyll/minima)
+[github.com/jekyll/minima](https://github.com/jekyll/minima)
 
 You can find the source code for Jekyll at
-{% include icon-github.html username="jekyll" %} /
-[jekyll](https://github.com/jekyll/jekyll)
+[github.com/jekyll/jekyll](https://github.com/jekyll/jekyll)
```
ashmaroli: It doesn't make a difference, but please replace this with text from Jekyll 3.5's `about.md` [template](https://raw.githubusercontent.com/jekyll/jekyll/v3.5.0/lib/site_template/about.md)

assets/icons.svg
```diff
@@ -0,0 +1,27 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+
+<symbol id="dribbble" fill-rule="evenodd" clip-rule="evenodd" stroke-linejoin="round" stroke-miterlimit="1.414"><path d="M8 16c-4.41 0-8-3.59-8-8s3.59-8 8-8 8 3.59 8 8-3.59 8-8 8zm6.747-6.905c-.234-.074-2.115-.635-4.257-.292.894 2.456 1.258 4.456 1.328 4.872 1.533-1.037 2.624-2.68 2.93-4.58zM10.67 14.3c-.102-.6-.5-2.688-1.46-5.18l-.044.014C5.312 10.477 3.93 13.15 3.806 13.4c1.158.905 2.614 1.444 4.194 1.444.947 0 1.85-.194 2.67-.543zm-7.747-1.72c.155-.266 2.03-3.37 5.555-4.51.09-.03.18-.056.27-.08-.173-.39-.36-.778-.555-1.16-3.413 1.02-6.723.977-7.023.97l-.003.208c0 1.755.665 3.358 1.756 4.57zM1.31 6.61c.307.005 3.122.017 6.318-.832-1.132-2.012-2.353-3.705-2.533-3.952-1.912.902-3.34 2.664-3.784 4.785zM6.4 1.368c.188.253 1.43 1.943 2.548 4 2.43-.91 3.46-2.293 3.582-2.468C11.323 1.827 9.736 1.176 8 1.176c-.55 0-1.087.066-1.6.19zm6.89 2.322c-.145.194-1.29 1.662-3.816 2.694.16.325.31.656.453.99.05.117.1.235.147.352 2.274-.286 4.533.172 4.758.22-.015-1.613-.59-3.094-1.543-4.257z"/></symbol>
```
ashmaroli: you can clean up further by removing `symbol` attributes other than `id`.
Its better to have `viewBox` attribute defined per symbol. (Note: reconfirm if the resulting SVG paths are still aligned in the browser after you make the changes)

_includes/social.html
```diff
@@ -0,0 +1,14 @@
+<ul class="social-media-list">
+  {% if site.dribbble_username %}<li><a href="https://dribbble.com/{{ site.dribbble_username | cgi_escape | escape }}"><svg class="svg-icon" xmlns="http://www.w3.org/2000/svg"><use xlink:href="/assets/icons.svg#dribbble"></use></svg> <span class="username">{{ site.dribbble_username | escape }}</span></a></li>{% endif %}
```
ashawley: This I can do

about.md
ashawley: Ok, I'll look in to changing the about page to be in-line with jekyll.

assets/icons.svg
ashawley: I'm not sure the consequences of this.  This wasn't brought up in the previous review of the original PR.  Can someone confirm?  Otherwise, we can probably do a follow-up after this is merged.

ashmaroli: I couldn't find a valid reference pointing to the dire consequences of this, but irrespective of the decision, if you want a working demo you can checkout #146

ashawley: To clarify, I don't know SVG enough to know the consequence of the change, and the original poster, @domingohui, is out of radio contact to defend the original use of `symbol`.

ashmaroli: okay..

ashmaroli:
Apart from one other place you need to make an edit, this looks okay to me.
/cc @DirtyF Requesting a review from you.. Thanks.

assets/icons.svg
DirtyF: @ashawley Don't bother to manually optimize the SVG, that's whats [SVGO](https://github.com/svg/svgo) is for. No problem is you prefer we deal with this in another PR.

about.md
```diff
@@ -4,12 +4,15 @@ title: About
 permalink: /about/
 ---
 
-This is the base Jekyll theme. You can find out more info about customizing your Jekyll theme, as well as basic Jekyll usage documentation at [jekyllrb.com](http://jekyllrb.com/)
+This is the base Jekyll theme. You can find out more info about customizing your Jekyll theme, as well as basic Jekyll usage documentation at [jekyllrb.com](https://jekyllrb.com/)
 
-You can find the source code for the Jekyll new theme at:
-{% include icon-github.html username="jekyll" %} /
+You can find the source code for Minima at GitHub:
+[jekyll][jekyll-organization] /
```
DirtyF: Should we keep the include approach here to be able to use something like `{% include social.html site="github" %}`?

_config.yml
```diff
@@ -8,6 +8,15 @@ description: > # this means to ignore newlines until "baseurl:"
...
 github_username:  jekyll
+rss: rss
+dribbble_username: jekyll
+facebook_username: jekyll
+flickr_username: jekyll
+instagram_username: jekyll
+linkedin_username: jekyll
+pinterest_username: jekyll
+youtube_username: jekyll
+googleplus_username: +jekyll
```
DirtyF: I would rather only keep existing Jekyll accounts and leave other services commented to avoid unnecessary 404s.

_includes/social.html
```diff
@@ -0,0 +1,13 @@
...
+  {% if site.twitter_username %}<li><a href="https://www.twitter.com/{{ site.twitter_username | cgi_escape | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#twitter"></use></svg> <span class="username">{{ site.twitter_username | escape }}</span></a></li>{% endif %}
+  {% if site.youtube_username %}<li><a href="https://youtube.com/{{ site.youtube_username | cgi_escape | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#youtube"></use></svg> <span class="username">{{ site.youtube_username | escape }}</span></a></li>{% endif %}
+  {% if site.googleplus_username %}<li><a href="https://plus.google.com/{{ site.googleplus_username | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#googleplus"></use></svg> <span class="username">{{ site.googleplus_username | escape }}</span></a></li>{% endif %}
+  {% if site.rss %}<li><a href="{{ site.baseurl }}/feed.xml"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#rss"></use></svg> <span>{{ site.rss | escape }}</span></a></li>{% endif %}
+</ul>
```
DirtyF: Should we rather loop on a `social` array in the config to apply DRY principle here?

about.md
ashmaroli: That approach was changed upstream to allow users switch themes via GH-Pages.

But this about.md is specific to Minima.
It also serves as a showcase page for what Minima looks like  when used with `jekyll new --`

Confusion.. confusion..

_includes/social.html
ashmaroli: That'd be a breaking change.. which is being covered (differently albeit) in #146

_config.yml
ashawley: Good catch. I agree. I had done that originally, but it got reverted while testing.  Commenting them out again...

_includes/social.html
ashawley: Right, I think the point of this PR is to just add new sites, but not break the existing ones, Twitter and GitHub.

ashawley:
Are there any other corrections?  It seemed like there was a lot of enthusiasm in the original PR for improving the social links in minima.

petermattia:
Hi, 
New Jekyll user here. I know Jekyll is quite popular with academics/grad students. Could you also add a Google Scholar link here? Thanks!

ashawley:
I would think adding more social links wold be possible, but the hope (at least mine is) is that this initiative is on it's way to getting merged and then adding other social sites can be taken up later.  The original work on this PR began almost 6 months ago so hopefully improved social links get incorporated.

ashawley:
So the SVG files used to be an include file, but this PR makes it an asset file.  Is there a risk of name collisions with an asset called `icons.svg`?  Should it be something like `assets/minima/social-icons.svg` instead?

ashmaroli:
> Is there a risk of name collisions with an asset called `icons.svg`?

Good call.
There's definitely a risk of collision if the user already has an `assets/icons.svg` at their source.
But if this PR is going to released as Minima 3.0, then we're covered to a certain extent. Existing sites are locked to `~> 2.0`

ashmaroli:
Requesting a review from you, @benbalter 
Thanks

_includes/social.html
```diff
@@ -0,0 +1,13 @@
+<ul class="social-media-list">
+  {% if site.dribbble_username %}<li><a href="https://dribbble.com/{{ site.dribbble_username | cgi_escape | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#dribbble"></use></svg> <span class="username">{{ site.dribbble_username | escape }}</span></a></li>{% endif %}
+  {% if site.facebook_username %}<li><a href="https://www.facebook.com/{{ site.facebook_username | cgi_escape | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#facebook"></use></svg> <span class="username">{{ site.facebook_username | escape }}</span></a></li>{% endif %}
+  {% if site.flickr_username %}<li><a href="https://www.flickr.com/photos/{{ site.flickr_username | cgi_escape | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#flickr"></use></svg> <span class="username">{{ site.flickr_username | escape }}</span></a></li>{% endif %}
+  {% if site.github_username %}<li><a href="https://github.com/{{ site.github_username | cgi_escape | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#github"></use></svg> <span class="username">{{ site.github_username | escape }}</span></a></li>{% endif %}
```
benbalter: Should this default to `site.github.owner_name` if not set?

_includes/social.html
```diff
@@ -0,0 +1,13 @@
...
+  {% if site.twitter_username %}<li><a href="https://www.twitter.com/{{ site.twitter_username | cgi_escape | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#twitter"></use></svg> <span class="username">{{ site.twitter_username | escape }}</span></a></li>{% endif %}
+  {% if site.youtube_username %}<li><a href="https://youtube.com/{{ site.youtube_username | cgi_escape | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#youtube"></use></svg> <span class="username">{{ site.youtube_username | escape }}</span></a></li>{% endif %}
+  {% if site.googleplus_username %}<li><a href="https://plus.google.com/{{ site.googleplus_username | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#googleplus"></use></svg> <span class="username">{{ site.googleplus_username | escape }}</span></a></li>{% endif %}
+  {% if site.rss %}<li><a href="{{ site.baseurl }}/feed.xml"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#rss"></use></svg> <span>{{ site.rss | escape }}</span></a></li>{% endif %}
```
benbalter: If JekyllFeed were used, it'd be nice to be able to set this automatically.

benbalter: Also, we should use the relative URL filter which can take into account misuses of baseurl.

_includes/social.html
benbalter: Rather than introduce 10 new top-level config flags that we know we want to break, I'd rather we implement the social hash described in #146, and have a fallback to retain support for the old `foo_username` syntax for the two existing ones.

_includes/social.html
```diff
@@ -0,0 +1,13 @@
+<ul class="social-media-list">
```
benbalter: I really wish we could have a hash of site => domain and then could iterate through e.g.,

```liquid
%{ for site_hash in site.social_sites %}
  {% if site.social[site_hash[0]] %}
    <li>
      <a href="https://{{ site_hash[1] }}/{{ site.social[site_hash[0]] | cgi_escape | escape }}">
        <svg class="svg-icon"><use xlink:href="/assets/icons.svg#{{ site_hash[0] }}"></use></svg> 
        <span class="username">{{  site.social[site_hash[0]] | escape }}</span>
      </a>
    </li>
  {% endif %} 
{% endfor %}
```

ashmaroli: I :heart: this idea

_includes/social.html
ashawley: That makes sense.  Is that possible?  What would it take to do?  Just a squirrelly if-conditional?

_includes/social.html
ashmaroli: The problem with this idea though is it doesn't validate the hash keys. Any key defined under the `site_hash` would be passed to the for block even if Minima doesn't have the svg icon for it..

_includes/social.html
ashmaroli:
```
"{{ 'feed.xml' | relative_url }}"
```

_includes/social.html
ashawley: Even if you could, the URL escaping rules aren't consistent for the sites.

_includes/social.html
ashawley: Is that an upstream Jekyll thing?

_includes/social.html
ashmaroli: say the hash is assigned to variable called `socials`, then I think the following will do fine:
```
{% assign github_user = social.github | default: site.github_username %}
<li><a>{{ github_user }}</a></li>
```

_includes/social.html
ashmaroli: Additional functionality provided by [`jekyll-github-metadata`](https://github.com/jekyll/github-metadata) plugin

_includes/social.html
```diff
@@ -9,5 +9,5 @@
...
   {% if site.social_sites.youtube %}<li><a href="https://youtube.com/{{ site.social_sites.youtube | cgi_escape | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#youtube"></use></svg> <span class="username">{{ site.social_sites.youtube | escape }}</span></a></li>{% endif %}
   {% if site.social_sites.googleplus %}<li><a href="https://plus.google.com/{{ site.social_sites.googleplus | escape }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#googleplus"></use></svg> <span class="username">{{ site.social_sites.googleplus | escape }}</span></a></li>{% endif %}
-  {% if site.rss %}<li><a href="{{ site.baseurl }}/feed.xml"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#rss"></use></svg> <span>{{ site.rss | escape }}</span></a></li>{% endif %}
+  {% if site.rss %}<li><a href="{{ 'feed.xml' | relative_url  }}"><svg class="svg-icon"><use xlink:href="/assets/icons.svg#rss"></use></svg> <span>{{ site.rss | escape }}</span></a></li>{% endif %}
```
ashmaroli: is there 2 whitespaces after `_url` ? Can't really tell..

ashawley: Yes, good catch.  Amended the commit

_includes/social.html
ashawley: Ok, it's using `relative_url` now

_includes/social.html
ashawley: Got it, thanks @ashmaroli

ashawley:
> Existing sites are locked to `~> 2.0`

Sure, but what about when they upgrade?

ashmaroli:
An upgrade to the next major version is a conscious decision.
A major version bump in a project that follows semver usually means breaking change. Its on the user to read the changelog or release notes and make an informed decision whether to upgrade.

_includes/social.html
benbalter: Ahh, I don't know the best way to do it, but ideally, Minima would maintain the mapping of known services to their domains and would loop through that list looking for a corresponding key in the site's config.

ashmaroli: > Minima would maintain the mapping of known services to their domains

If Minima is ready to maintain a mapping, then it seems doable.

ashawley:
As it stands,

1. An icon for Google Scholar was requested. It should be possible, but was deferred to a follow-up PR
1. The suggestion to make a parameterized map on the include template was considered friendly, but isn't possible at the moment because
   1. of issues with Jekyll and liquid, but also because 
   1. the quoting/escaping rules are different for each of the social sites.
1. It was agreed to work on optimizing the SVG in a follow-up PR
1. I've just now renamed the `icons.svg` to `minima/social-icons.svg` to avoid the risk of colliding with someone who already has an `icons.svg` file under `assets/`

ashmaroli:
> I've just now renamed the `icons.svg` to `minima/social-icons.svg` to avoid the risk of colliding..

Why not just rename it to `minima-social.icons.svg`? I don't see the need for a subdirectory with just one file..

ashawley:
> I don't see the need for a subdirectory with just one file..

Sure, thought the subdirectory would be a good thing to having going forward.  Happy to rename, though

DirtyF:
@ashawley Could you post a screenshot or point to a demo? Thanks.

ashmaroli:
> thought the subdirectory would be a good thing

it would only force users to maintain an `assets/minma/` if they need to make a change to our assets.. lets introduce subdirectories when reeaally required for better organisation.. :wink:

ashawley:
@DirtyF Here's a demo http://jekyll-minima-ashawley.s3-website-us-east-1.amazonaws.com/

ashawley:
> lets introduce subdirectories when reeaally required for better organisation..

Ok, agreed.  I've removed the subdirectory

DirtyF:
@ashawley thanks for setting this up, a quick check shows that we could do better on small screens:

minima-social-small-screen

ashawley:
> Maybe use display-inlne:block, flexbox or css-grid with @support here to show icons on columns?

Not sure how to do that.  Happy to accept PRs on my fork from you and I'll merge.  Maybe easier to just improve how it looks in a follow-up PR?

ashawley:
Thanks for the Minima theme and Jekyll generally.  

The real thanks go to @domingohui for the actual work of making this PR.  I just tried to get it across the finish line.

ashawley:
I've rebased with the latest master commits and added an entry to the release history file.

Could this get merged, soon?  People are requesting other social links, but this is blocking those questions.

The last status report was [Oct 25, 2017](https://github.com/jekyll/minima/pull/141#issuecomment-339373832)

PR #142, opened 2017-07-13
Docs: clarification about page title

antgel:
I had to dig into the source code to find out why my page links weren't automagically generated - this patch will help others in the same boat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Refers to snippets of code within the `_includes` directory that can be inserted
   - `footer.html` &mdash; Defines the site's footer section.
   - `google-analytics.html` &mdash; Inserts Google Analytics module (active only in production environment).
   - `head.html` &mdash; Code-block that defines the `<head></head>` in *default* layout.
-  - `header.html` &mdash; Defines the site's main header section.
+  - `header.html` &mdash; Defines the site's main header section. By default, pages with a defined `title` attribute will have links displayed here.
   - `icon-* files` &mdash; Inserts github and twitter ids with respective icons.
 
 ### Sass
--- a/404.html
+++ b/404.html
@@ -1,0 +1,24 @@
+---
+layout: default
+---
+
+<style type="text/css" media="screen">
+  .container {
+    margin: 10px auto;
+    max-width: 600px;
+    text-align: center;
+  }
+  h1 {
+    margin: 30px 0;
+    font-size: 4em;
+    line-height: 1;
+    letter-spacing: -1px;
+  }
+</style>
+
+<div class="container">
+  <h1>404</h1>
+
+  <p><strong>Page not found :(</strong></p>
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -35,7 +35,7 @@ Refers to files within the `_layouts` directory, that define the markup for your
 
   - `default.html` &mdash; The base layout that lays the foundation for subsequent layouts. The derived layouts inject their contents into this file at the line that says ` {{ content }} ` and are linked to this file via [FrontMatter](https://jekyllrb.com/docs/frontmatter/) declaration `layout: default`.
   - `home.html` &mdash; The layout for your landing-page / home-page / index-page.
-  - `page.html` &mdash; The layout for your documents that contain FrontMatter, but are not posts.
+  - `page.html` &mdash; The layout for your documents that contain FrontMatter, but are not posts. By default (unlike posts), page FrontMatter requires a `title` attribute, in order for a link to the page to be generated (in `header.html`).
```
ashmaroli: IMO, instead of adding it here, it'd be better next to **`_includes - header.html`** with the following edit:
```diff
- By default (unlike posts), page FrontMatter requires a `title` attribute, in order for a link to the page to
- be generated (in `header.html`).
+ Displays a list of Pages with `title` set in the FrontMatter
```

antgel: I see where you're coming from, but that doesn't address part of the underlying issue, that no title in Page FrontMatter, leads to... nothing, whereas with a Post, no title DWIM. Do you think your version is explicit enough that it's clear that *no* Page title => *no* output?

ashmaroli:
> Do you think your version is explicit enough that it's clear that no Page title => no output?

Perhaps.. perhaps not.
I'm open for further edits. But, I recommend that it be a short n simple sentence.

The trouble with
```
By default (unlike posts), page FrontMatter requires a `title` attribute, in order for a link to
the page to be generated (in `header.html`).
```
is that it doesn't convey the essence in a simple manner. I'll break it out to illustrate.

---
`By default` is used for configurable items. For example,
```
By default the color is red. But you can override it by defining the '$accentColor' variable
```

---

`(unlke posts)` `=>`

posts or any other collection items have no say here..in the `header.html`. Its just `Pages` and Pages alone. Unnecessary confusion.

`(in header.html)` `=>`
won't be necessary if the note is included in the right place

antgel:
@ashmaroli Thanks for the feedback, commit amended to suit, let me know what you think.

ashmaroli:
LGTM now.

/cc @DirtyF 
Should we also make a note about `header_links` configuration (under `Usage`) as part of this PR? :thinking:

DirtyF:
Thank you @antgel for the edit!  😄  🔗  💥 

@jekyllbot merge +docs

benbalter:
@jekyllbot: merge +docs

Thanks @antgel!

PR #147, opened 2017-07-22
Style feed in sync with theme

ashmaroli:
Making an attempt to style feeds in sync with other pages using Minima

feeds

/cc @pathawks @benbalter


DirtyF:
minima relies now on jekyll-feed, this is nice but we don't want to maintain a style for the feed as t's mostly intented to be read by external apps. Thanks for the effort.

PR #150, opened 2017-07-27
Added pagination to home.html

aterenin:
This PR adds pagination support to Minima.

To enable, set `paginate: 5` in `_config.yml` and run Jekyll.

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -6,18 +6,17 @@ layout: default
 
   {{ content }}
 
-  {% if site.paginate %}
-  {% assign posts = paginator.posts %}
-  {% else %}
   {% assign posts = site.posts %}
+  {% if site.paginate %}
+    {% assign posts = paginator.posts %}
   {% endif %}
 
   <h1 class="page-heading">Posts</h1>
 
   <ul class="post-list">
-    {% assign date_format = site.minima.date_format | default: "%b %-d, %Y" %}
     {% for post in posts %}
       <li>
+        {% assign date_format = site.minima.date_format | default: "%b %-d, %Y" %}
         <span class="post-meta">{{ post.date | date: date_format }}</span>
 
         <h2>
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -4,6 +4,8 @@ layout: default
 
 <div class="home">
 
+  {{ content }}
+
   {% if site.paginate %}
   {% assign posts = paginator.posts %}
   {% else %}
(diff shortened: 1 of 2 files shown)


ashmaroli:
Thank you for your contribution, @aterenin 
Any associated styling for the pagination?

IMO, it'd be better if the code for pagination be abstracted to an `include`. That way it could easily used in another layout as well..

How about some addition to the documentation? It'd be better to note that only `jekyll-paginate` is supported by GitHub Pages. The successor `jekyll-paginate-v2` isn't..

_layouts/home.html
```diff
@@ -6,12 +6,18 @@
...
+  {% else %}
+  {% assign posts = site.posts %}
+  {% endif %}
+
   <h1 class="page-heading">Posts</h1>
-  
+
   <ul class="post-list">
-    {% for post in site.posts %}
+    {% assign date_format = site.minima.date_format | default: "%b %-d, %Y" %}
```
ashmaroli: let's handle moving this `assign` in another PR..

_layouts/home.html
```diff
@@ -6,12 +6,18 @@
 
   {{ content }}
 
+  {% if site.paginate %}
+  {% assign posts = paginator.posts %}
+  {% else %}
+  {% assign posts = site.posts %}
+  {% endif %}
```
ashmaroli: Would the following work?
```
{% assign posts = site.posts %}
{% if site.paginate %}
  {% assign posts = paginator.posts %}
{% endif %}
```

aterenin:
Requested changes have been made.

I am not sure what the right way to abstract it would be. There's essentially two parts: one which generates a list of posts, and one which generates `Page 1 of 5`. I do not see how the same code would be used elsewhere. Is there a good example I may not have considered?

I'm not sure precisely how Github Pages handles pagination. In my own site, which uses this code, I do not need to enable any plugins for pagination - just adding the line `paginate: 5` suffices. I'll update the readme file to indicate this.

pathawks:
I think this is handled with the **jekyll-paginate** plugin, which [GitHub Pages includes](https://pages.github.com/versions/).

aterenin:
Wasn't aware GH pages includes that - it isn't in my `_config.yml` but I presume since the code works that it still gets loaded.

Regarding the two separate plugins - `jekyll-paginate` and `jekyll-paginate-v2` - my code should seamlessly support both, as it checks for existence of a `paginator` object rather than a specific plugin. I added a link to the readme to point toward the relevant Jekyll page, since many people will be using GH pages.

Issue #140, opened 2017-07-06
Obliterate Whitespace


ashmaroli:
Once #139 merges, Minima will depend on Jekyll 3.5 and that means, we can safely make the switch to Liquid 4 within our templates.

Most notably, use the new `whitespace-control` feature to remove unnecessary whitespace in the generated HTML output.

PR #146, opened 2017-07-20
Convert social-links column into a row of icons

ashmaroli:
**Continuation of #141 and #112**
**Milestone:** `3.0.0`

Breaking Changes:
  - All the social-links are clubbed under the namespace `minima.social_links`.
  - Footer layout has been changed along with the associated markup 

socials

PR #153, opened 2017-08-11
removes unwanted whitespace after icons

CSchoel:
At the end of the SVG files for the social media icons there is a newline character that will be included in the `<span>` tag in the template. Firefox renders this as an additional space character that adds more whitespace between logo and text than what is indicated by the CSS file. I noticed this, when I tried to add a PNG icon for Research Gate that did not align properly with the other icons.

In reference to #140, I think the correct solution for this small display bug is to use the `whitespace-control` feature of Liquid 4.

I also increased the padding used for the whitespace in the CSS, so that the visual impression stays (almost) the same.


CSchoel:
I just noticed that `whitespace-control` did not really fix the problem. I was experimenting with different options and did not notice that I also had removed the whitespace inside the actual SVG files when I checked if the whitesprace-control tags would help. It seems that they only remove additional whitespace generated by liquid-tags and not whitespace within included files.

So the solution really seems to be to alter the SVG-files. Another option could be to switch to a different tag other than `<span>`, but I did not want to change the HTML itself.

ashmaroli:
The entire social-icons sections are being considered for an overhaul. See #141 and #146 for proposals..

Also, I couldn't reproduce the whitespace bug on Firefox by comparing the [footer at current version on GitHub Pages](https://jekyll.github.io/minima/)

Hence if possible, can you post a screenshot of the said bug at your end? If you're able to post an additional screenshot  of any other browser on your system for comparison, all the better..

CSchoel:
The bug is quite subtle and probably not that relevant if the social icons will change anyway.

Still, for the record I made some screenshots in Firefox and Chrome (from the Link to the gh-pages version that you posted) that show the issue:

* In Firefox you can actually see the text content as separate element in the development tools. The total width of the span element should be 16 pixels from the SVG + 5 pixels padding, but the whitespace adds another 4 pixels of content.

    minima-space
* In Chrome the issue is the same, but it is a little less visible. Again, the toal width is 25 instead of 21 pixels.

    minima-space_chrome

ashmaroli:
Golly, I see it now.
Though I still don't think that a newline character would have a visible presence in HTML.

CSchoel:
You are right, newlines should not matter in HTML. That also got me confused at first glance, but the nasty detail is that this newline is added between two `<span>` tags. Because of this, it is condensed into a single space character like any other whitespace inside a node's text content.

This

```
<span>no</span><span>space</span>
```

renders differently than this

```
<span>no</span>
<span>space</span>
```

ashmaroli:
Cool! Thank you @CSchoel  for unearthing the bug-source.
The latest `master` is now requiring a minimum of Jekyll 3.5, which means Liquid 4.0 access to all from next release.
And there are active proposals to do away with `includ-ing` social icons.. So this bug might get resolved automatically.

DirtyF:
Just tested with latest version and spaces are fine when I use svg-icon class on an image in Firefox.

PR #165, opened 2017-10-24
Support for non-default Jekyll Feed path

tommyang:
Jekyll Feed has an option for [non-default path](https://github.com/jekyll/jekyll-feed#already-have-a-feed-path).
If specified, it should be used for the feed link in head.html instead of the default path.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -4,7 +4,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   {% seo %}
   <link rel="stylesheet" href="{{ "/assets/main.css" | relative_url }}">
-  <link rel="alternate" type="application/rss+xml" title="{{ site.title | escape }}" href="{{ "/feed.xml" | relative_url }}">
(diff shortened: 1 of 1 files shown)


pathawks:
I wonder if this could be combined with #129 to only include the feed link if `site.feed.path` is set? 😍

pathawks:
Oh! And thanks for submitting your first pull request to this project, @tommyang! :tada::beers:

tommyang:
Thanks!

Oops. I was not previously aware of that pull request.

> only include the feed link if site.feed.path is set

Unfortunately, I don't think that would work since `jekyll-feed` [does not _require_](https://github.com/jekyll/jekyll-feed/blob/f879ed22076fce4cff98b22fec7d0ec85f9a3199/lib/jekyll-feed/generator.rb#L24) `site.feed.path` to be set.

DirtyF:
Closing this as we now rely on jekyll-feed.

PR #175, opened 2018-01-03
Improve the wording in index.md

adamatan:
The current comment in index.md was confusing to me as a new Jekyll developer; it seemed to suggest that the content should be added directly to _layouts/home.html instead of index.md. This fix strives to clarify the roles of each file.


ashmaroli:
The `index.md` here is only used for the demo Minima, on GitHub Pages. For the clarification to benefit new developers on Jekyll, please submit a patch the template at [`jekyll/jekyll`](https://github.com/jekyll/jekyll/blob/master/lib/site_template/index.md) as well..

adamatan:
@ashmaroli What should be the right reference to the HTML file? "Change the `_layouts/home.html` file in your theme directory?

ashmaroli:
> What should be the right reference to the HTML file?

The best thing would be to make changes to Jekyll's official documentation.

The `index.md` you're editing here will only be available for those whose are cloning this repo. In that situation, a simple `Edit _layouts/home.html to make changes to the layout for this page` would be sufficient.
But for users starting fresh by running `jekyll new`, they'll have to follow the instructions in the section at the URL mentioned in the comment, to customize the layout.. i.e. copy `_layouts/home.html` from the Minima gem (or this repo) into their site source directory and then edit that copy.

PR #177, opened 2018-01-04
Honor DNT for Google Analytics

MikkCZ:


MikkCZ:
@DirtyF r?

MikkCZ:
@pathawks I am not sure if a template should enforce this. IMO it should definitely be configurable. The decision is on the project maintainers. If the decision is made, I will be more than happy to add the config entry to enable it by default.

pathawks:
I am asking you why any site would not want this. If there isn’t a good reason, why add a configuration option?

MikkCZ:
If you think so, I can update the PR of course. Should I?

MikkCZ:
I have updated the patch.

DirtyF:
@jekyllbot: merge +minor

DirtyF:
@jekyllbot: merge +minor

PR #179, opened 2018-01-14
fix(css): prevent title from getting covered

theetrain:
Good day to you. When using minima for the first time today, I noticed my site's long title can get covered by the mobile hamburger menu button. I thought I'd help.

## Before

image

## After

image


DirtyF:
Note: While adding some padding and adjusting line-height is OK for the smallest breakpoint, we'll  have to deal with other edge cases like a few links in the nav with a longer title.

super-long-title

theetrain:
@ashmaroli thanks for your suggestion. I've addressed the line height and pushed my amended commit.

@DirtyF I may be able to address this concern in a separate PR. I have a couple ideas:

## 1. Move the nav elements before the title

image

## 2. Write a liquid template to count the number of characters in the title and nav links to decide whether to show the hamburger or not

**header.html**
```html
<div class="wrapper">
    ...

    {% if page_paths %}
                   <!-- if title is too long, always show mobile nav -->
      <nav class="site-nav{% if site.title.size > 45 %} touch{% endif %}">
        <input type="checkbox" id="nav-trigger" class="nav-trigger" />
```

There may be more edge cases to consider, such as having many nav links wrap to a new line.

PR #180, opened 2018-01-14
Cache includes that do not depend on page data

ashmaroli:
Includes such has `header.html` and `footer.html` are prime candidates for being cached:
  - they are **page-agnostic** &mdash; they're the same for every page, for every url (where they're rendered..)
  - they employ expensive Liquid `for` loop computation
`{% for item in list %} do something {% endfor %}`

This cache-ing can be made possible by using the [`jekyll-include-cache`](https://github.com/benbalter/jekyll-include-cache) launched in December 2017

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -5,7 +5,7 @@
 
   <body>
 
-    {% include header.html %}
+    {% include_cached header.html %}
 
     <main class="page-content" aria-label="Content">
       <div class="wrapper">
@@ -13,7 +13,7 @@
       </div>
     </main>
 
-    {% include footer.html %}
+    {% include_cached footer.html %}
 
   </body>
 
--- a/minima.gemspec
+++ b/minima.gemspec
@@ -17,6 +17,7 @@ Gem::Specification.new do |spec|
   end
 
   spec.add_runtime_dependency "jekyll", "~> 3.5"
+  spec.add_runtime_dependency "jekyll-include-cache", "~> 0.1"
   spec.add_runtime_dependency "jekyll-seo-tag", "~> 2.1"
   spec.add_development_dependency "bundler", "~> 1.12"
 end


DirtyF:
@ashmaroli I like the idea of promoting this useful plugin, but I wonder what @benbalter thinks of adding an unsupported plugin by GitHub Pages to Jekyll's default theme.

ashmaroli:
@DirtyF Thank you for mentioning Ben. Would love to know if he has any plans to bundle this plugin with `github-pages` in the near future :smiley:

benbalter:
> I like the idea of promoting this useful plugin, but I wonder what @benbalter thinks of adding an unsupported plugin by GitHub Pages to Jekyll's default theme.

I'm 👎 on the idea of adding Jekyll Include Cache to the *default* theme (unless someone else wants to adopt and make a significant investment in its user experience). It is not currently *supported* by GitHub Pages, although, IIRC, it is whitelisted if you are able to implement it on your own.

Jekyll include cache is great for speeding up large and complex sites if you know what you're doing, but the user experience has some rough edges with the potential to silently bite the user.

I think @ashmaroli OP is a great example of the complexity. Users would have to understand the concept of a page-agnostic include (or page-specific content), and especially on a site with multiple authors, adding "a small feature" to an include, could potentially break things in a very confusing way.

If someone (or the community) would like to adopt Jekyll Include Cache and round out the rough edges, dramatically improve the documentation, provide build time feedback, etc., I'm all for it, but to be honest, I'm not in a position to personally support the plugin's use by inexperienced users (it currently has me and one other watcher), especially if it becomes part of the default theme (often the first step to using Jekyll or static site generators for the very first time).

**Edit**: To clarify, it's technically solid, it's just that the user experience for non-advanced users is significantly lacking.

ashmaroli:
closing as the plugin is not suited for a *starter-theme*..

PR #189, opened 2018-01-27
Update social_sites in README

AndrewJudson:
This changes the documentation to match the version in _config.yml
in https://github.com/jekyll/minima/pull/141.


ashawley:
I realized that this was actually an unintentional defect and rolled back `social_sites` in #190.  So if you switched to using `social_sites`, Andrew, you will need to go back to just `twitter_username` and `github_username` and similar for the new sites.

I wouldn't have caught on to this unless you had filed this documentation correction, so thanks again for catching this.

PR #190, opened 2018-01-27
Revert 2a5bcb0 by removing social_sites hash

ashawley:
Introducing a `social_sites` hash was an unintentional change I committed and pushed but wasn't caught during the review.  The consequence is that users who upgrade will silently have missing social links when their existing `_config.yml` is using `twitter_username` or `github_username`.  Only the rss feed will remain.

I apologize for this defect.  It was a local experiment that was intended to be made on an experimental local git branch but was accidentally committed during code review comments and was subsequently never caught in the review.

Hopefully, this change can be reverted as a bug fix release 2.2.1 posthaste.


DirtyF:
@ashawley Thanks. [v2.2.1 was just released on rubygems](https://rubygems.org/gems/minima).

PR #198, opened 2018-02-05
Adding Mastodon to the social networks

Bandie:
With this you can add a Mastodon username+instance to your config which will show it nicely like all the other social services.


parkr:
What does this look like when the options are specified? What if a user is on multiple instances?

Bandie:
It looks like this:

jekyll_mst
You don't see the instance name intentionally because
1. it would break the design
2. Mastodon does it the same for known users within a conversation

---
I assumed that someone would add just one instance.
If you request it to have more than one we'd have to change it.
We could do something like
```
mastodon_username1: foo1
mastodon_instance1: bar1
mastodon_username2: foo2
...
```
…if you don't have a better idea.

parkr:
@Bandie Thanks for the screenshot! Hard to test when I'm on mobile ☎️ . 😄 

As for multiple instances, I see lots of folks using multiple instances so if we're going to support this, I'd love to support multiple instances right off the bat. What do you think about this in the `_config.yml` file? 

```yaml
mastadon:
- username: parkr
  instance: foo.bar.com
```

Then in the Liquid, you'd check for `site.mastadon` in your `if`, and then `for` loop over `site.mastadon`.

Bandie:
Thanks @parkr, now I know that the html stuff is called Liquid. :D
Here you are~

_config.yml
```diff
@@ -19,6 +19,14 @@ rss: rss
 #youtube_username: jekyll
 #googleplus_username: +jekyll
 
+# Mastodon instances
+# mastodon:
+# - username: jekyll
+#   instance: mastodoninstance.tld
```
pathawks: How about using `example.com` here?

Bandie: Agreed™

parkr:
@jekyllbot: merge +minor

PR #199, opened 2018-02-06
fix: styling main element in IE 11

mmktomato:
The styling about `.page-content` is not affected in IE 11 because the default value of `display` of `main` element is 'inline'. (It's 'block' in Chrome 64.)
It will be fixed by adding 'display: block;' to `main` element.


DirtyF:
@jekyllbot: :shipit:

DirtyF:
@jekyllbot: 🚢 +minor

parkr:
@jekyllbot: merge +fix

mmktomato:
Thx for reviewing and merging !

PR #183, opened 2018-01-16
Fix link to feed

jekyllbot:
fix #98

From the start, minima was meant to be a minimal blog theme with posts and a link to a feed. 
This PR aims only to fix the actual link to the feed.

PR #201, opened 2018-02-20
Social icons should resolve a site's baseurl properly

### LINKED ISSUE #197: Social icon path should include the baseurl if set
It looks like the social icons are hardcoded to /assets/minima-social-icons.svg. However, if the baseurl is set, it needs to be pre-pended to that path, otherwise the svg will not be found.

ashmaroli:
Fixes #197


ashawley:
While fixing the images, fix the XML feed so it's site relative, as well?

ashawley:
Unless the XML feed is _already_ site-relative.

ashmaroli:
> fix the XML feed so it's site relative

it already is.. please comment on the diff if it isn't..

DirtyF:
@jekyllbot: :shipit: +fix

pathawks:
@ashawley What XML feed?

pathawks:
@ashawley I don't mean to beat a dead horse, but can you point to that spot in the code?

I don't follow Minima as closely as others, but I don't think Minima includes its own feed, so it shouldn't have any link to `feed.xml` in the template.

ashawley:
@pathawks It's in the output:

```
$ grep -e '/feed.xml' ./_site/index.html 
  <link type="application/atom+xml" rel="alternate" href="/feed.xml" title="Your awesome title" />
    <p class="rss-subscribe">subscribe <a href="/feed.xml">via RSS</a></p>
  <li><a href="/feed.xml"><svg class="svg-icon"><use xlink:href="/assets/minima-social-icons.svg#rss"></use></svg> <span>rss</span></a></li>
```

pathawks:
Oh oh oh! I follow.

In #183 we removed the hardcoded feed link to use the **jekyll-feed** plugin instead. Thanks for explaining :beers:

itsjavi:
any chance to have this released soon? perhaps as `v2.3.1`? 
thx

PR #204, opened 2018-03-07
Fix HTML Validator Errors

### LINKED ISSUE #75: role=banner attribute on <header>
The `<header>` tag includes a `role=banner` attribute. Since `<header>`'s implicit ARIA semantics is actually `role=banner` and `<header>` is not used as a descendant of either an `<article>` or a `<section>` element in the theme (see the [ARIA documentation](https://www.w3.org/TR/html-aria/#docconformance)), I think the role can be removed.

ankitrgadiya:
- [x] Warning: The banner role is unnecessary for element header
- [ ] Error: Attribute href not allowed on element data at this point
- [ ] Error: Element data is missing required attribute value

Resolve #75 #211


ankitrgadiya:
The other 2 errors are due to the following line in [_includes/footer.html](https://github.com/jekyll/minima/blob/master/_includes/footer.html#L2)

```
<data class="u-url" href="{{ "/" | relative_url }}"></data>
```

I don't exactly know what is the use of this line.

PR #205, opened 2018-03-07
Add better system fonts

yashghelani:
Add a system font stack to support different OSes for a more comfortable & pleasing look. This exact stack is used by https://github.com/ as well


DirtyF:
@jekyllbot: :shipit:

PR #208, opened 2018-03-09
Make `social` configuration section as more legibly

kacperduras:
Probably clearer and prettier :D


_config.yml
```diff
@@ -7,25 +7,24 @@ description: > # this means to ignore newlines until "twitter_username:"
...
-# Mastodon instances
-# mastodon:
-# - username: jekyll
-#   instance: example.com
-# - username: jekyll2
-#   instance: example.com
+social:
+  # mastodon:
+  # - username: jekyll
+  # instance: example.com
```
pathawks: `instance` should be indented two more spaces.

kacperduras:
```
However, this would be a major change, and every existing site using Minima’s social icons would need to change their config.
```
@pathawks Right, I didn't think about it. What's your idea for this? Because I have only one - adding support for the older configuration while retaining my PR.

_config.yml
kacperduras: Fixed: e80011158bcce15c099c9b7927a967ec8e59d035

pathawks:
That’s a good question.

Using the old style as a `default` seems reasonable to me, so long as the maintainers of this theme agree that this is worth doing.

/cc: @DirtyF @ashmorali

ashmaroli:
related-comments; https://github.com/jekyll/minima/pull/112#issuecomment-304941832 onwards..
prior-art: #146

P.S. I'm not a maintainer of this theme..

ashawley:
Maybe there is a way to support both the existing keys and a new `social` prefix?  That might make the code really squirrelly, though.

I'm tempted to just support Ashwin's existing PR to just get it right in 3.0.

kacperduras:
```
Maybe there is a way to support both the existing keys and a new social prefix?
```

@ashawley I wrote about this above. If necessary, I can add such a commit :)

I suggest you wait for feedback from @DirtyF

DirtyF:
What @pathawks said. Let's not change current keys for now.

We can revisit this in 3.0 with proper comments in the configuration file.
Related:  We could avoid to add a `rss` entry in this array and display rss icon is `feed.xml` is generated (which is the case by default).

```
# Link to your social profiles
social:
  twitter: jekyll
  github: jekyll
  # linkedin: 
  # pinterest
  ....
```

ashawley:
I suggest closing in favor of #146.

DirtyF:
Closing in favor of #146.

PR #217, opened 2018-03-26
Reintroduce removed social includes for backwards compatibility

### LINKED ISSUE #215: A file was included in `/_layouts/default.html` that is a symlink or does not exist in your `_includes` directory
My github page, that built successfully over 2 weeks ago, now is failing with the following error:

> The page build failed for the `master` branch with the following error:
> 
> A file was included in `/_layouts/default.html` that is a symlink or does not exist in your `_includes` directory. For more information, see https://help.github.com/articles/page-build-failed-file-is-a-symlink/.
> 
> For information on troubleshooting Jekyll see:
> 
>   https://help.github.com/articles/troubleshooting-jekyll-builds

Could you suggest how to resolve this?
Do I need to update something or add some files?
Is it a bug in minima theme?

ashmaroli:
Fix #215 

Since users *may* have a `footer.html` at their source that still links to these removed includes **and** because Jekyll simply aborts if it can't find the requested include.


ashmaroli:
I recommend this be shipped and included in the `github-pages` gem a.s.a.p

ashawley:
The consequence of this change is that people with custom footer.html will not get the new social links.  That's commendable.

This seems like a lot of flip-flopping.  Hopefully, this doesn't risk making more confusion, rather than less.  This avoids broken builds, but could also produce users who ask "Why can't I use the new social links?"

So will these includes be deleted in minima 3.0?

DirtyF:
@ashawley The most important is to no break builds for the thousands of people that were using minima. It's fair to say that when we ship 3.0, we will document the upgrade and remove those.

ashawley:
> The most important is not to break builds for the thousands of people that are using minima

I'm just trying to discuss the change so we can review it.  

I'm sure there are "thousands of people" using minima, but how many people have sourced the existing footer.html file?

DirtyF:
No matter how many people customized their footer, we don't want to break their site. We'll be more cautious next time and use post-install message to notify for breaking changes.

ashawley:
There appear to be about 30,000 public users of minima on Github alone:

https://github.com/search?q=minima+language%3AYAML&type=Code

But only a dozen people who have a footer.html referencing the old social files.

https://github.com/search?q=icon-twitter+path%3A_includes%2Ffooter+language%3AHTML&type=Code

mateuszbuda:
@ashawley You only count public repos.

ashawley:
> You only count public repos.

Yes, I meant to qualify that and forgot.  Thanks.

nathanKramer:
FWIW this broke my build and was pretty frustrating trying to work out why without much help from jekyll's error message:

> "Failure: A file was included in `/_layouts/default.html` that is a symlink or does not exist in your `_includes` directory."

I wasn't in your search @ashawley despite having a public repo so your number is probably low.

ashmaroli:
@nathanKramer *This PR* is a ***fix*** for the issue you mentioned.. Did you update to Minima-2.4.1? If you were simply referencing a past experience, then ignore this comment..

ashawley:
Is the github-pages gem going to get updated with this fix?

https://github.com/github/pages-gem/releases

PR #223, opened 2018-03-30
Stick footer for short posts

### LINKED ISSUE #222: Footer should be at the bottom of the screen
For long posts, I have to scroll down to see the footer, and the page ends right after -- there is no unwanted space below the footer. This is good.

For short posts, the footer is right below the post and there is a lot of space below the footer, between it and the end of the screen. This is bad. The footer should be at the bottom of the screen in this case.

For a default theme, there are a lot of style issues... but I am learning a lot about CSS :)

DirtyF:
Fix #222 

minima-stick-footer

### DIFF
--- a/_posts/2016-05-19-super-short-article.md
+++ b/_posts/2016-05-19-super-short-article.md
@@ -1,0 +1,7 @@
+---
+layout: post
+title: "Some articles are just so short that we have to make the footer stick"
+categories: misc
+---
+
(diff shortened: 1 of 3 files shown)


ashmaroli:
LGTM! :+1:

montyf:
Why not apply `flex: 1;` to `.post-content` as well?

I just tried this solution and it worked for my blog! It's much simpler than the previous solution I had and doesn't require knowing the size of the footer beforehand.

parkr:
Nice! I have never seen the flex properties. Are they cross-browser compatible?

DirtyF:
@parkr https://caniuse.com/#feat=flexbox

PR #228, opened 2018-04-10
Update config

### LINKED ISSUE #227: RSS Feed is broken
I recently deploy a Gihub Pages with minima theme. Everything works perfect except the RSS Feed.
Click on "subscribe via RSS" will show an 404 page. 
I also check minima demo page. The feed is broken also.

screen shot 2018-04-10 at 3 42 55 pm

DirtyF:
Link to feed is broken on https://jekyll.github.io/minima/ I guess it's because we have to tell in the config that the theme uses the `jekyll-feed` plugin.

Also listed `jekyll-seo-tag` to be consistent

- [x] remove useless directives as we already have defaults defined for markdown engine and excluded files.

fix #227

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -40,8 +40,12 @@ minima:
 # - about.html
 
 # Build settings
-markdown: kramdown
 theme: minima
+
+plugins:
+ - jekyll-feed
+ - jekyll-seo-tag
+
 exclude:
   - Gemfile
   - Gemfile.lock
--- a/_config.yml
+++ b/_config.yml
@@ -45,7 +45,3 @@ theme: minima
 plugins:
  - jekyll-feed
  - jekyll-seo-tag
-
-exclude:
-  - Gemfile
-  - Gemfile.lock


ashmaroli:
I think you can delete the `exclude:` array as well since Jekyll has got a more wholesome array by default..
Though I recommend that you move the deletions out of this PR and into a dedicated branch..

ashmaroli:
I wonder how GitHub Pages built the site successfully without loading the plugins..
Liquid would've aborted with an `Unknown Tag {% seo %}` (or `{% feed_meta %}`) exception..

parkr:
> I wonder how GitHub Pages built the site successfully
GitHub Pages [defaults to `jekyll-theme-primer`](https://github.com/github/pages-gem/blob/master/lib/github-pages/configuration.rb#L19) if you don't specify a theme, and it [depends on `jekyll-seo-tag`](https://github.com/pages-themes/primer/blob/master/jekyll-theme-primer.gemspec#L18), so it's automatically required by Jekyll now. 😸

Issue #149, opened 2017-07-25
Make minima a Progressive Web App


DirtyF:
Following discussion on #145 

[Progressive Web Apps](https://developers.google.com/web/progressive-web-apps/)  aims at providing a better user experience on the web by providing offline support, home screen button and push notifications for already mobile-friendly websites. Offline support and push notifications are provided through a Service Worker (a standard JavaScript API). 

As GitHub pages offers HTTPS support, it's possible to enhance a responsive website and turn it into a PWA. All you need is to write [some generic JavaScript and some Liquid](https://jamesiv.es/jekyll/amp/2017/05/09/serviceworkers-with-jekyll.html) to generate the `sw.js` file. A `manifest.json`file will be used for the home screen button. You can also imagine pushing a notification for a new post (eg. a new Jekyll release on jekyllrb.com 😄 )

In order to do so, a theme has to provide:

- HTTPS (GitHub Pages does that)
- some favicons and a [manifest.json](https://w3c.github.io/manifest/) file 
- a offline.html page (very much like a 404.html)
- a service worker, the `sw.js` file could be generic or customizable (as you have to list the file you want the browser to serve offline, e.g. assets, the last posts, some pages, etc. You also have to generate a cache version (through a timestamp for instance) to invalidate the previous cache.

If we want to allow this on GitHub Pages, we cannot rely on a gem like [jekyll-offline](https://github.com/jeremiak/jekyll-offline) which I tested a while back and it wasn't satisfying.

pathawks:
Why is including this in a theme better than making it available as a plugin?

DirtyF:
@pathawks It isn't. 😄  

If you prefer we could break this issue in two parts and just rely on a quality plugin (that doesn't exist yet) for the offline and cache invalidation part.

ashawley:
Franks says the jekyll-offline gem "wasn't satisfying" and a plugin would be better, but would a plugin be possible?  As written, #234 wouldn't be of much use to someone using minima as a Jekyll theme.

Issue #145, opened 2017-07-20
Minima 3.0


ashmaroli:
I think we should start development on a version 3.0.
  - Convert our liquid templates into Liquid 4.0 (mainly to employ the `Whitespace Control` feature.)
  - Because of above, our dependencies will change:
    - `"jekyll", "~> 3.5", ">= 3.5.1"`
    - `"liquid", "~> 4.0"`
  - Redesign the social links in the footer into a horizontal section at the end of `footer` ( #146 )
  - Introduce any other breaking change here.
  - Continue maintenance of version 2.x

/cc @jekyll/minima

pathawks:
> Continue maintenance of version 2.x

Is there enough developer momentum to maintain two branches of Minima?

ashmaroli:
> Is there enough developer momentum to maintain two branches of Minima?

that kinda went right above my head...
are you saying there aren't enough maintainers for the two branches?, or
are you saying there isn't a need to start dev on v3.0 now?

parkr:
If there are any outstanding issues on 2.x then I'd get them pushed out on 2.x now, then start converting.

I'm 👍  to shipping a new major version, e.g. v3.x, which requires Liquid 4.

Strangehill:
- I would like to see the css be mobile-first. I gave this a shot in pr #130 a while back so it might need revisiting.
- I'd also like to see styles in minima split into smaller modules, to better encourage piecemeal customization of the gem theme. What I think probably happens now, the way gem themes work, is that updating the minima theme on a site is rarely done. Whenever styles in a large file are tinkered with the site owner is from then on forced to manually bring in any updates to the gem theme found in those customized files.
- How throwing in a favicon, a manifest.json, and maybe a service worker?

ashmaroli:
> maybe a service worker

what would this do?

DirtyF:
> manifest.json, and maybe a service worker

Ideally offline support and PWA should be available for all themes not just minima. Currently I don't know of any reliable plugin, but you can use an npm package like [sw-toolbox](https://www.npmjs.com/package/sw-toolbox).

pathawks:
> I would like to see the css be mobile-first.

Buzzwords aside, what does this entail?

parkr:
[...]

PR #234, opened 2018-04-24
Add a Service Worker and an offline page

chrisfinazzo:
Based on the discussion in #149, I took another crack at this. As the aim of Minima is to show what unmodified Jekyll looks like (E.g, `jekyll start`), I cached files based on the contents of `lib/site_template`. 

For a 3.0 release (#145), this seems like a reasonable thing to include.


pathawks:
I am not sure that this is in keeping with the spirit of having a _minimal_ theme.

chrisfinazzo:
@pathawks,

It probably isn't (in a strict sense), but this seems like a lighter-weight solution compared to using a module like Workbox. Although I've used that on a small site and it does the job well, it would require people to deal with NPM. That isn't difficult, but it is a bit heavy in terms of dependencies.

As previously mentioned, this is probably something for 3.0, so I'm happy to wait on this for a while.

DirtyF:
Thanks @chrisfinazzo I'd really much like if we could provide a better user experience and allow offline browsing for minima. Shouldn't we add a timestamp to the `staticCacheName` here?

DirtyF:
@chrisfinazzo we don't want any `npm` dependency to stay _minimal_. Thanks for using the native API.

DirtyF:
[Going offline](https://abookapart.com/products/going-offline) by @adactio is out today, let's read it and make sure we have a solid strategy here 😉

chrisfinazzo:
@DirtyF I read it this  afternoon, so this was on my mind 😃.

To your point regarding a timestamp on the `CacheName`, if we update the version constant, that _should_ cause a new registration. To be honest, the section on caching is a bit dense, so I probably need to spend a bit more time thinking about what the right thing to do is.

DirtyF:
@chrisfinazzo  I was thinking of something like `const staticCacheName = "version-{{site.time | date: '%Y%m%d%H%M%S'}}";`

chrisfinazzo:
@DirtyF Ah, I see now. Will update.

ashawley:
This would be a nice addition.  Thanks for working on adding it.

1. The `offline.html` is a useful template, but wouldn't be shipped with the gem-based theme.
2. The same for `sw.js`.  It would need to be located in `assets/js` or some such.
3. Maybe call the file `offline.js` instead of `sw.js`?
4. If the service worker is added on every page, and the `sw.js` exists, but the `offline.html` does not, the service worker fetch will fail.  This is a Javascript console error.  Loading the site while offline will fail to load the offline page ⚡️, but the browser knows how to degrade gracefully to just a regular "This site can't be reached", so 👍.
5. It would be nice to have the offline page cache be configurable, but it seems like this should be a Jekyll plugin?

sw.js
```diff
@@ -0,0 +1,61 @@
...
+const staticCacheName = "version-{{site.time | date: '%Y%m%d%H%M%S'}}";
+
+addEventListener('install', installEvent => {
+     console.log('The service worker is installing...');
+     skipWaiting();
+     installEvent.waitUntil(
+       caches.open(staticCacheName)
+       .then( staticCache => {
+         return staticCache.addAll([
+           '/assets/css/style.css',
```
ashawley: The name of the stylesheet was moved to `/assets/main.css`.

sw.js
```diff
@@ -0,0 +1,61 @@
...
+     })
+   );
+});
+
+
+addEventListener('fetch', fetchEvent => {
+     console.log('The service worker is listening.');
+     const request = fetchEvent.request;
+     fetchEvent.respondWith(
+       caches.match(staticCacheName)
```
ashawley: This wasn't retrieving assets from the cache when offline, because it needs to be changed to `caches.match(request)`.

chrisfinazzo:
@ashawley,

Thanks for the feedback. I have a branch in my local repo that has some of the changes you mentioned, I will review there and then add those to this PR.

sw.js
chrisfinazzo: Last I touched this, it had been [changed][] in order to match Pages' "house style". 

[changed]: https://github.com/jekyll/minima/pull/305/commits/7a64517c183fc4a823ebfee599806fd9ce509284

ashawley: Ah, ok, so I had it backwards.  It got moved to `assets/css/style.css`.  That explains it.

chrisfinazzo:
> 4. If the service worker is added on every page, and the `sw.js` exists, but the `offline.html` does not, the service worker fetch will fail.  This is a Javascript console error.  Loading the site while offline will fail to load the offline page ⚡️, but the browser knows how to degrade gracefully to just a regular "This site can't be reached", so 👍.


In GitHub Pages environments, the custom `404.html` is used. In terms of fallbacks, it's not much work to modify an `.htaccess` or `nginx.conf` to point to it - if your host doesn't do this behind the scenes already.

ashawley:
I had suggested:

> 2. The same for `sw.js`.  It would need to be located in `assets/js` or some such.

According to [Google's primer on Service Workers](https://developers.google.com/web/fundamentals/primers/service-workers/) that wouldn't really work.

> One subtlety with the `register()` method is the location of the service worker file. You'll notice in this case that the service worker file is at the root of the domain. This means that the service worker's scope will be the entire origin. In other words, this service worker will receive fetch events for everything on this domain. If we register the service worker file at `/example/sw.js`, then the service worker would only see fetch events for pages whose URL starts with `/example/` (i.e. `/example/page1/`, `/example/page2/`).

It seems it would need to live at the root directory.

> 3. Maybe call the file `offline.js` instead of `sw.js`?

This name is cryptic, but I suppose it is the standard name for this kind of thing.

PR #235, opened 2018-04-25
Remove Internet Explorer 8/9 tag

adunning:
Setting `X-UA-Compatible` to `IE=edge` only seems relevant for preventing quirks mode in IE 9 and earlier; see <https://stackoverflow.com/questions/6771258/what-does-meta-http-equiv-x-ua-compatible-content-ie-edge-do> for an explanation.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -1,6 +1,5 @@
 <head>
   <meta charset="utf-8">
-  <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   {%- seo -%}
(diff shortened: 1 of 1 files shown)


DirtyF:
Thanks @adunning for bringing that up. We currently align on [H5BP](https://github.com/h5bp/html5-boilerplate/blob/master/dist/index.html#L5) here. I believe it's also relevant for IE10. 

We don't heavily test on IE browsers, a browserstack screenshot shows that it degrades gracefully, so I don't mind keeping the meta for now.

win8_ie_10 0

adunning:
Thanks for the explanation! IE 10 (and later) browsers understand it, but as far as I understand its quirks mode changed the IE 9 behaviour, making the tag irrelevant.

PR #236, opened 2018-04-25
Fix SVG validation error

adunning:
Attribute "xmlns" is not allowed to appear in element "symbol".


ashawley:
Funny trivia is that the stackoverflow icon isn't supported by Minima, because it was [dropped for technical reasons](https://github.com/jekyll/minima/pull/141#issuecomment-316476405) at the last minute.  Its place in the SVG file was preserved in case it was revisited.  I can't explain why it didn't get the `xmlns` trimming that the others got.  Regardless, I'm sure the stackoverflow part needs to fixed to validate the entire file.

PR #237, opened 2018-04-28
Add Bing webmaster validation meta tag

BerkhanBerkdemir:
Hello,

I added this meta tag because I wanted to keep DNS records simple and clean.

Thank you.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -8,4 +8,7 @@
   {%- if jekyll.environment == 'production' and site.google_analytics -%}
     {%- include google-analytics.html -%}
   {%- endif -%}
+  {%- if jekyll.environment == 'production' and site.bing_webmaster -%}
+    <meta name="msvalidate.01" content="{{ site.bing_webmaster }}">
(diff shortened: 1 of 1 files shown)


chrisfinazzo:
👋🏻 @BerkhanBerkdemir, good to have you here.

Can we generalize this so that any search verification scheme could be added? I suspect people will want to add others for completeness sake (Google, Yahoo, etc.).

If you prefer, I can open another PR for those.

desiredpersona:
Webmaster verification is fully supported via [jekyll-seo-tag](https://github.com/jekyll/jekyll-seo-tag/blob/master/docs/usage.md).

```
webmaster_verifications:
  google: 1234
  bing: 1234
  alexa: 1234
  yandex: 1234
  baidu: 1234
```

chrisfinazzo:
@desiredpersona Oy, now I get it. Was thinking people would have to add their own key value pairs under the `site` namespace, but already done.

Carry on...

PR #245, opened 2018-05-17
Make rss optional on home

c-martinez:


chrisfinazzo:
Hi @c-martinez, welcome to the project.

I'm curious, as most sites have a feed (either using Jekyll's official plugin or made from scratch), what's the rationale on making this optional? Particularly on a homepage where most people will want to subscribe, it seems counterintuitive.

c-martinez:
Hi @chrisfinazzo, thanks

I guess my reasoning was mostly consistency -- twitter and github are optional so why not rss? Plus rss seems to already be optional in [social.html](https://github.com/jekyll/minima/blob/master/_includes/social.html`) so it made sense to me.

DirtyF:
rss is optional in the social icons because there's already a link in the main area, so that's only for DRY purpose. other items are optional in social because we have no way of knowing which social meda accounts they have. RSS is enable by default. 

If you wanna change the home layout, just override the theme.

Issue #243, opened 2018-05-10
tables can create horizontal overflow on mobile


decathorpe:
This makes some of my posts look (and behave, because of the possible horizontal scrolling) badly on mobile.

Example: https://decathorpe.com/2018/04/08/this-week-in-fedora-2018w14.html

Zooming out to see the whole page at once looks even worse.

First image: Top of one of my posts (table not fully visible)

![1](https://user-images.githubusercontent.com/12836693/39872002-e7585642-5466-11e8-99ce-da8e2b7aa4a6.png)

Second image: Scrolled to the right (rest of table visible, but all other styling is off, for example the top bar and navigation menu)

![2](https://user-images.githubusercontent.com/12836693/39872006-ea7c2fc4-5466-11e8-837b-d409b5240a4d.png)

M0nica:
Would any of these responsive table implementations be ideal? A scrollbar appears for the entire table on smaller screens.
[http://maxdesign.com.au/articles/simple-responsive-table/](http://maxdesign.com.au/articles/simple-responsive-table/ )

DirtyF:
@M0nica That's what we do on jekyllrb.com, care to submit a fix?

M0nica:
Sure @DirtyF I will work on this.

M0nica:
I submitted a pull request [https://github.com/jekyll/minima/pull/246](https://github.com/jekyll/minima/pull/246) however, I believe this change can only be implemented if tables are updated to be HTML based instead of Markdown-based. That is how the Jekyll site approaches this.

decathorpe:
Hm. I just noticed that over-wide code blocks *do* have the scrollable behavior I am looking for. Would it be possible to do the same thing for tables?

PR #246, opened 2018-05-20
Update Tables to Show Scrollbars on Mobile

M0nica:
I updated Minima to cause scrollbars to appear on tables on mobile devices/small screens if the screen size is too small to display the entire table. This is to address the problem of tables not overflow and distorting the mobile experience which was mentioned in [https://github.com/jekyll/minima/issues/243](https://github.com/jekyll/minima/issues/243).

 I set it up to follow the same conventions as [https://github.com/jekyll/jekyll](https://github.com/jekyll/jekyll).  However, in order to implement this functionality, users have to use HTML instead of Markdown for their tables since the CSS styling cannot be added without a specific div class being defined around every table. Is this trade-off okay?

### DIFF
--- a/_posts/2016-05-20-this-post-demonstrates-post-content-styles.md
+++ b/_posts/2016-05-20-this-post-demonstrates-post-content-styles.md
@@ -89,17 +89,81 @@ Proin eget nibh a massa vestibulum pretium. Suspendisse eu nisl a ante aliquet b
 
 ### Tables
 
-Title 1               | Title 2               | Title 3               | Title 4
---------------------- | --------------------- | --------------------- | ---------------------
-lorem                 | lorem ipsum           | lorem ipsum dolor     | lorem ipsum dolor sit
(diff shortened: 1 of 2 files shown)


DirtyF:
Shouldn't we target all tables in posts here instead of adding a class in order to support Markdown tables?

M0nica:
@DirtyF I need to see if there's a way to implement the desired functionality by only targeting the table element. The examples I saw and how it is implemented on Jekyll's site is by adding the CSS styling to a div around the table.

DirtyF:
@M0nica yeah, we use HTML tables in Jekyll docs for configuration options, but here the need might be different. Thanks for looking at this.

BerkhanBerkdemir:
Hey @DirtyF, will they implement this feature because I have same issue/feature request.

DirtyF:
Closing in favor of #296

Issue #264, opened 2018-08-08
Page masthead image above/next to/instead of site title?


uliwitness:
Minima is a gorgeously simple theme, but given most sites have some sort of logo or masthead image, I would think allowing for an (optional) masthead image to go with the site title would be an addition that doesn't yet constitute feature-creep. I'm thinking of something where you specify a `masthead-image`, `masthead-left-image` or `masthead-above-image` in the `_config.yml`.

Would anyone be interested in adding this? If I managed to somehow add this feature to my copy of Minima, would there be interest in merging that back into the mainline?

uliwitness:
Okay, I managed to build this. In `_includes/header.html`, replace the existing title output code with:

```yaml
	{%- if site.masthead -%}
	<a href="{{ "/" | relative_url }}"><img class="site-masthead" src="{{ site.masthead | relative_url }}" alt="{{ site.title | escape }}" /></a>
	{%- else -%}
	<a class="site-title" rel="author" href="{{ "/" | relative_url }}">{{ site.title | escape }}</a>
	{%- endif -%}
```

In `_sass/minima/_layout.scss`, add the following CSS class:

```css
.site-masthead {
  display: block;
  margin-left: auto;
  margin-right: auto;
  padding-top: $spacing-unit / 2;
  padding-bottom: $spacing-unit / 2;
}
```

And in `_config.yml` add:

```yml
# Specify the name of a file to show as a masthead graphic at the top of each page, instead of a title, here:
# masthead: /my_masthead.jpg
```

Would be grateful if someone could put that into the official Gem.

DirtyF:
Would you mind opening a PR so that we can test this out?

uliwitness:
See PR #266.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

PR #266, opened 2018-08-08
Add logo image (#264)

uliwitness:
Can now specify the name of an image file in _config.yml that will be shown instead of the site title on every page.


DirtyF:
Closing as we could add background options for headers in minima 3 if #294 gets approved.

uliwitness:
Isn't a background for a header a different use case compared to a logo image?

PR #270, opened 2018-08-10
add favicon

derjoker:
Optional: we can keep the code but remove the favicon.


DirtyF:
What problem does this solve?

derjoker:
Normally I would expect a favicon for my website. Would it be better if the user can add his own favicon by just copying (and replacing) `favicon.ico` in the root?

DirtyF:
@derjoker do you have a source file for the favicon to export it in other format?

derjoker:
@DirtyF no, I just generated this from [favicon.io](https://favicon.io/), where we can also have .png format.

derjoker:
> I'm not sure we even need to declare the link in the head, as browsers automatically check for this default favicon.

@DirtyF Didn't know that before ;-P I tried, and yes, the browsers recognize `favicon.ico` automatically, then it doesn't make sense to add the link in the head any more.

To help users get a nice favicon, either we add a comment somewhere in README, or we simply include a `favicon.ico` by default. Which one do you prefer?

By the way, here's the favicon you asked for ;-)

screen shot 2018-09-03 at 11 06 39 pm

DirtyF:
> To help users get a nice favicon, either we add a comment somewhere in README, or we simply include a favicon.ico by default. Which one do you prefer?

We can do both, add the default favicon to avoid the current 404 error and mention favicon.io for instance to replace the default one.

DirtyF:
closing in favor of #278, thanks @derjoker for bringing that matter to our attention.

Issue #259, opened 2018-07-29
Social network icons are misaligned with text


varkor:
All the social network icons appear vertically misaligned, as in:
![image](https://user-images.githubusercontent.com/3943692/43371777-8f132732-9390-11e8-829c-41c3218fdc2b.png)

I think this line should be changed to: `vertical-align: middle;`:
https://github.com/jekyll/minima/blob/be3a0a16c27c9154b6c9543ca864ae8472d55934/_sass/minima/_base.scss#L217

PanderMusubi:
Yes, see also this screenshot:

![2018-08-14-154859_1920x1080_scrot](https://user-images.githubusercontent.com/198964/44095769-e8634de4-9fd9-11e8-9bd3-66c497ea1e56.png)

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

varkor:
This appears to be fixed in master.

PR #271, opened 2018-08-15
Vertically aligned middle for social icons. (_base.scss/.svg-icon)

intern0t:
This particular fix is referenced to the issue #259.

I basically changed `vertical-align: text-top;` to `vertical-align: middle;` to align the icons properly.

Working Demo

Feel free to report if this doesn't fix, I'll try and fix them.


DirtyF:
social-icons

intern0t:
@DirtyF, I think when we remove `vertical-align` property completely, it's is not centering the icons but rather baselining them. As you can see the base of the `icon` and the `span` are the same when you removed the `vertical-align` property as a whole. `middle` argument in `vertical-align` centers it well.

In addition, `text-bottom` value for `vertical-align` property makes it look vertically centered as well.

PR #281, opened 2018-09-11
Migrate from analytics.js to gtag.js

mladenilic:
Use [gtag.js](https://developers.google.com/analytics/devguides/collection/gtagjs/) instead of [analytics.js](https://developers.google.com/analytics/devguides/collection/analyticsjs/) snippet, as it is now the preferred option.

Google Analytics docs suggest using `window['ga-disable-GA_TRACKING_ID'] = true;` as a way to disable tracking, so that's added as well. :)

Here's the link to the docs: https://developers.google.com/analytics/devguides/collection/gtagjs/user-opt-out

Cheers!


DirtyF:
>  as it is now the preferred option

@mladenilic could you provide official source that supports this statement?

mladenilic:
@DirtyF there's no official statement that I'm aware of, it's just that when you create the Google Analytics account, the tracking code provided is the new gtag.js one.

https://s3.amazonaws.com/drei.one-public/gtag.png

PR #286, opened 2018-10-15
Load favicon

jonchoi:
The current branch doesn't seem to load the favicon for me (Mac OS X,
Chrome).

Using this solution fixed it: https://stackoverflow.com/questions/30551501/unable-to-set-favicon-using-jekyll-and-github-pages/30552322

#270

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -1,6 +1,8 @@
 <!DOCTYPE html>
 <html lang="{{ page.lang | default: site.lang | default: "en" }}">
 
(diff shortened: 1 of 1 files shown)


DirtyF:
@jonchoi thanks, this has already been fixed in #278

DirtyF:
Hi @jonchoi 

Even if present in a theme favicon.ico don't get copied to destination, so adding this won't  be enough.
As assets folder get copied, you'll have to place your favicons in assets directory and point to the custom path in the meta.

DirtyF:
Closing in favor of #287

PR #289, opened 2018-10-17
Hide nav toggle if nav is empty.

samueldiethelm:
Nav toggle button would display in mobile view even if nav dropdown is empty.


_includes/header.html
```diff
@@ -3,9 +3,10 @@
   <div class="wrapper">
     {%- assign default_paths = site.pages | map: "path" -%}
     {%- assign page_paths = site.header_pages | default: default_paths -%}
+    {%- assign titles_size = site.pages | map: 'title' | join: '' | size -%}
```
DirtyF: What if a page without title is added in `header_pages`?

DirtyF:
Please expose your use case first.

_includes/header.html
samueldiethelm: It would be joined as an empty string and would not affect the the joined string size `titles_size.` Either way, as the code is right now, a page with no title would not show in the nav either since you have `{%- if my_page.title -%}` further down in the code, which I have also changed to check the joined string size `titles_size`.

samueldiethelm:
> Please expose your use case first.

If you use the theme purely as a blog, and the only _page_ you have is the home page which displays your posts, and which has no title, then `page_paths` will not be empty, but in the loop the title `my_page.title` of the pages will be empty. Therefore the menu toggle will be visible (on mobile) but with no elements in the dropdown when you click on it.

PR #308, opened 2018-11-05
Added rel=me to Mastodon

sphinxc0re:
I added a `rel="nofollow me"` to the social links since it declares the links as personal ones. Very practical especially for mastodon since it uses this to verify a personal link.


ashawley:
Does it only need to be added to the Mastodon link, then, or all social links?

Do you have any links to docs that could explain the reason for this change?

pathawks:
Why `nofollow`?

sphinxc0re:
@ashawley No, for my personal needs, Mastodon would suffice. But I thought it might be good to just go all the way for the social links.

@pathawks I actually just looked at how GitHub generates links to your homepage when you add one to your profile:

screenshot 2018-11-05 at 22 07 43

That said, after a quick Google I found out that this tells search engines not to count the linking into the rank of the linked-to site. I can remove that, if it isn't what we want. Since this is a backlink to something, I don't know if `nofollow` has any purpose 🤷‍♂️

ashawley:
Ok, I think changing all the social links is arbitrary, then.  That doesn't seem necessary, then, since no one else has asked for it.

I'm still trying to understand why Mastodon is asking you to do this for verification.  Do you have any docs you are following about that?

sphinxc0re:
Of course: https://docs.joinmastodon.org/usage/basics/#link-verification

DirtyF:
Let's take a look at HTML spec:

> The nofollow keyword indicates that the link is not endorsed by the original author or publisher of the page, or that the link to the referenced document was included primarily because of a commercial relationship between people affiliated with the two pages. — https://html.spec.whatwg.org/multipage/links.html#link-type-nofollow

I don't think you wouldn't endorsed your own social profile.

sphinxc0re:
@DirtyF Thanks for the clarification and yes, you are right, I would endorse my own profile 😁

ashawley:
> Of course: https://docs.joinmastodon.org/usage/basics/#link-verification

Ah, thanks!

LGTM.  

PS. Should probably rename PR to "Added rel=me to Mastodon" or similar.

PR #311, opened 2018-11-12
Add Wikipedia english to social

### LINKED ISSUE #310: Add Wikipedia Page as an option link in the footer
Hello,

Would it be possible to add a new option to link a wikipedia page through a simple addition of the parameter in the _config.yml?

ashawley:
Fixes #310 by linking to English Wikipedia page, `en.wikipedia.org/wiki/User:`.

I balked at doing the global logo:

https://en.wikipedia.org/wiki/File:Wikipedia-logo-v2.svg

I'm not sure we could make a tiny grayscale version that was identifiable.

Instead, the logo is based on the public domain Wikipedia "W":

https://commons.wikimedia.org/wiki/File:Wikipedia%27s_W.svg

Social icons with Wikipedia


DirtyF:
That social icons are starting to become a burden, we should come with a better solution to allow users to add their favorite service, I'm thinking of removing all current icons alltogether in next major version.

DirtyF:
x

DirtyF:
HI, thanks for the submission, we don't intend to support any more services in this theme. PLease override your social.html locally to add support for desired services.

PR #313, opened 2018-11-17
Remove unused none display in _layout.scss

bangseongbeom:


DirtyF:
@jekyllbot: merge +fix

PR #316, opened 2018-11-29
Fix missing </header> bug.

Fenzland:
Add a new line at start of `_includes/header.html` so that `</header>` won't missing after run `jekyll build`.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -1,3 +1,4 @@
+
 <header class="site-header">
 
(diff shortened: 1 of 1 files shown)


DirtyF:
closing as `<header class="site-header">` is not missing on https://jekyll.github.io/minima/

Fenzland:
Sorry, bug is in one of plugin that I use, with a incorrect regex. minima is ok.

PR #319, opened 2018-12-06
Removing redundant title from footer.html

sbhat423:
Site title is present in `<h2>`. When site author is not mentioned there will be two site titles one below the other, which is redundant.


ashmaroli:
Thank you @sbhat423. Sorry for the delay in getting this merged.

PR #321, opened 2018-12-08
 Add `overflow-wrap: break-word` to body tag

bogdanvlviv:
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

combined

closes #306

### DIFF
--- a/_sass/minima/_base.scss
+++ b/_sass/minima/_base.scss
@@ -27,6 +27,7 @@ body {
   display: flex;
   min-height: 100vh;
   flex-direction: column;
+  overflow-wrap: break-word;
 }
 
 


DirtyF:
This looks like a duplicate of #306 

What problem are you solving?

bogdanvlviv:
Looks like almost the same, but I think we should apply `overflow-wrap: break-word` to the `body` since there could be a too long word on any part of a page, not only in a post title(https://github.com/jekyll/minima/pull/306).

Example:
[Too long word in the body of one of my article](https://bogdanvlviv.com/posts/ruby/rails/what-is-new-in-rails-6_0.html#activerecordrelationcreate_or_find_by-to-lean-on-unique-constraints) caused extra horizontal scroll before I [fixed](https://github.com/bogdanvlviv/bogdanvlviv.github.io/commit/408fda816c3c91083063e0a9ad959d7fdae2527d) it. So I decided to try to fix that directly for `minima`, not only for my website.

bogdanvlviv:
Could you please review this patch?

ashmaroli:
Thank you @bogdanvlviv and @DirtyF 
@jekyllbot: merge +bug

Issue #327, opened 2018-12-24
Add social icon for Goodreads


SkullTech:
Goodreads is the largest //best community for those who read books. I might take a stab at this, but I don't know anything about SVG, so I'm hoping someone else is generous enough to do this.

chrisfinazzo:
If anyone has a better source image than those available [here][], let me know. Using Imagemagick - and Potrace as an intermediary doesn't work well.

[here]: https://www.goodreads.com/about/press

SkullTech:
@chrisfinazzo Have you made any progress on this?

chrisfinazzo:
@SkullTech, I tried converting the "G" logo they have on their Press page, but that didn't turn out well - there's still a "round rect" in the final image, which isn't really what I had in mind. There is another logo there which has the full text, I'll see if that works better.

DirtyF:
FWIW here's an extraction of the G in SVG: https://www.dropbox.com/s/sf5ltcvs3ft44zi/goodreads.svg?dl=0

I was thinkin moving social profiles to a data file and let users handle this by themselves, we don't want to list 50 social services in our default config file.

chrisfinazzo:
🤔 

```
<symbol id="goodreads" fill-rule="evenodd" clip-rule="evenodd" stroke-linejoin="round" stroke-miterlimit="1.414"><path
      d="m35.245 3.194c-21.216 0-32.062 17.059-32.062 36.987 0 20.251 10.846 36.063 30.768 36.063 21.064 0 32.557-16.295 32.557-36.063.152-18.825-10.678-36.987-31.263-36.987zm31.415 53.268h-.315c-3.34 14.507-18.19 22.965-32.214 22.965-22.801 0-34.131-18.197-34.131-39.246 0-22.008 12.146-40.181 35.245-40.181 15.643 0 27.917 10.369 31.1 23.762h.315v-21.85h3.194v79.257c0 22.324-12.764 32.849-34.143 32.849-16.576 0-30.78-7.489-31.257-25.843h3.204c.642 16.273 13.079 22.65 27.896 22.65 19.787 0 31.106-9.402 31.106-29.656z"/></symbol>
```

<img width="101" alt="screen shot 2019-01-06 at 9 02 31 pm" src="https://user-images.githubusercontent.com/1534882/50745118-72f81a00-11f6-11e9-990c-f863f93c99f4.png">

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

[...]

PR #325, opened 2018-12-20
add gitlab support

chrisdecker1201:

PR #324, opened 2018-12-14
Add Kaggle to social

baleato:
I took the [logo](https://storage.googleapis.com/kaggle-competitions/kaggle/3136/media/kaggle-transparent.svg) from Kaggle's website ([Kaggle Logos section](https://www.kaggle.com/contact)). And edited with [SVG-Editor ](https://github.com/SVG-Edit/svgedit).

image

Closes #323

PR #333, opened 2019-01-10
Migrate social links to use a data file

chrisfinazzo:
Per @DirtyF's comment [here](https://github.com/jekyll/minima/issues/327#issuecomment-451765885).


_includes/social.html
```diff
@@ -1,17 +1,20 @@
...
-  {%- if site.microdotblog_username -%}<li><a rel="me" href="https://micro.blog/{{ site.microdotblog_username| cgi_escape | escape }}"><svg class="svg-icon grey"><use xlink:href="{{ '/assets/minima-social-icons.svg#microdotblog' | relative_url }}"></use></svg> <span class="username">{{ site.microdotblog_username| escape }}</span></a></li>{%- endif -%}
-  {%- if site.googleplus_username -%}<li><a href="https://plus.google.com/{{ site.googleplus_username| escape }}"><svg class="svg-icon grey"><use xlink:href="{{ '/assets/minima-social-icons.svg#googleplus' | relative_url }}"></use></svg> <span class="username">{{ site.googleplus_username| escape }}</span></a></li>{%- endif -%}
-  {%- if site.rss -%}<li><a href="{{ 'feed.xml' | relative_url }}"><svg class="svg-icon grey"><use xlink:href="{{ '/assets/minima-social-icons.svg#rss' | relative_url }}"></use></svg> <span>{{ site.rss | escape }}</span></a></li>{%- endif -%}
+{% for social in site.data.social %}
```
ashmaroli: This looping is wrong. `site.data.social` is not an array of items

ashmaroli:
@chrisfinazzo I would recommend that you model the change here based on #146 as following:
```
{%- assign social = site.data.social -%}
[...]
```
But then since data files inside theme-gems are not read-in by Jekyll, this will require additional changes to be a complete feature.

chrisfinazzo:
Oh, Ok, I see what you're getting at - will update. I knew the loop wasn't right, but at first thought I messed up something w.r.t Mastodon as it is nested inside the others. I have a similar setup on my own site, so that will be something to work off of.

DirtyF:
Maybe we could:

- only provide Github and Twitter by default in config.yml like it was before minima v2.2.0
- if a `social.yaml` file exist, use the items listed in the data file instead
- default social.yaml contains a few social media examples

Document how to add a social service: 
 - add an SVG icon
 - add entry in _data/social.yml

The main idea being minima only offer basic support for social links, and it's up to the user to add more services based on the given examples.

ashmaroli:
@DirtyF If #146 were to merge before this PR, we could simply have
```
{%- assign social = site.data.social | default: site.minima.social_links -%}
```
Either ways, things look like a breaking change..

chrisfinazzo:
@ashmaroli, Yeah, I noticed that we'd be stepping on each other's work here - either way you'd like for me to do this is fine. I'll close the other related PR's when this is done.

ashmaroli:
This needs some documentation now.
It should especially note that Minima will look **either within `site.data.social` or within `site.minima.social_links`**.

As of now, uncommenting an entry in the data file will cause Minima to simply ignore the entire listing in the config file.

ashmaroli:
@chrisfinazzo Please don't pull in other user's contributions into this PR.
When this PR gets merged, the commits will be squashed into one under your name.
The other user's contribution will be lost.

The proper way is to wait till this PR is merged and then help the other user resolve their conflicts or recommend starting a new PR with latest `master` branch.

chrisfinazzo:
> @chrisfinazzo Please don't pull in other user's contributions into this PR.
> When this PR gets merged, the commits will be squashed into one under your name.
> The other user's contribution will be lost.
> 
> The proper way is to wait till this PR is merged and then help the other user resolve their conflicts or recommend starting a new PR with latest `master` branch.

Yeah, this was my fault, I cherry picked that commit out of @baleato's repo and then updated it to match the format, which had changed since then. Merge didn't go right so I just grabbed the files and redid it, which isn't fair to @baleato as it doesn't really credit him properly.

There are a few similar outstanding PR's (#322, #325, #327), I'll ask them to get the latest master and update to the new format before we bring them in.

ashmaroli:
> I'll ask them to get the latest master and update to the new format before we bring them in.

That reminds me of another thing.
Minima was conceived to be a very minimal theme to get the get going with Jekyll and themes.
I'm not sure if adding every social media icon out there boards well with the "minimalistic" principle here.

Instead of adding icon-after-icon, we should simply convert our template into an extensible entity without hampering maintainability.

chrisfinazzo:
I still have the `_data` directory in my local copy and plan to actually do the migration when these few others close. IIRC (please, @DirtyF, correct me if I've got it backwards), doing it this way means people can turn on as many or as few of the services as they want. Not sure how I feel about asking people to write this themselves, even if it isn't all that difficult.

ashmaroli:
My vision is thus:
- convert the social-icon listing into an array of key-value pairs.
- support only a few popular social media
- ask the end-user to procure SVG and the domain format for the remaining social media
- Minima renders whatever icons they list in their config file / data file, if available in their `assets` directory

baleato:
Hi @chrisfinazzo, I've updated #324. I appreciate the effort. I'd suggest not to worry about the other change. I'll keep updating the PR in case of conflicts. That way I'll also get more insight on the latest changes ;)

chrisfinazzo:
> * ask the end-user to procure SVG and the domain format for the remaining social media

This is the key issue I think in determining whether or not the feature gets used. Prior [work][] on this shows that even if users are proficient enough to acquire an SVG for the service they want to use, it doesn't always come out right. This is usually because files are either unavailable in that format, or the XML used to create them is doing something which causes a display issue. 

[work]: https://github.com/jekyll/minima/issues/327#issuecomment-451765885

To that end, I would put this question to @jekyll/minima and @jekyll/stability to determine what the right way forward is.

For my part, I agree with your and @DirtyF's idea in principle, but wonder how much of it should be abstraced via `_data` or asking people to go the DIY route.

chrisfinazzo:
Closing this.

PR #334, opened 2019-01-13
Progress bar on Posts

IamLizu:
I have added a progress bar that will appear on pages having the _post_ layout. This could be easier for the visitors of a blog to know how much of the content they have read and how much left.


ashmaroli:
Thank you @IamLizu but this is a bit too much for a minimal theme such as Minima.

PR #338, opened 2019-01-17
Add pre-Safari 9.0 functionality in viewport tag

moajday:
After iOS 9 release, Safari browser zooms out everything automatically. However, suggested change into Minima theme viewport meta tag will fix the initial zoom problem.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -1,7 +1,7 @@
 <head>
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
-  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="viewport" content="width=device-width, initial-scale=1 shrink-to-fit=no">
   {%- seo -%}
   <link rel="stylesheet" href="{{ "/assets/css/style.css" | relative_url }}">
   {%- feed_meta -%}


DirtyF:
This is specific to Apple, I'd rather avoid adding this if we can.

Did you notice rendering problems? 
Do we have oversized elements on mobile that we need to style correctly? 
Do we apply `max-width` on images, and `overflow-x: scroll` on code blocks?

coliff:
`shrink-to-fit=no` is no longer needed. The default behavior was rectified with the release of iOS 9.3

Good research and explanation here: https://www.scottohara.me/blog/2018/12/11/shrink-to-fit.html

moajday:
> Did you notice rendering problems?
**No. I assumed it might be helpful for users who are using the particular version of iOS 9.**
> Do we have oversized elements on mobile that we need to style correctly?
**We do not have oversized elements.**
> Do we apply `max-width` on images, and `overflow-x: scroll` on code blocks?
**We do not need to apply these attributes either.**

@DirtyF You have already closed the pull. I'm deleting the branch.

moajday:
> `shrink-to-fit=no` is no longer needed. The default behavior was rectified with the release of iOS 9.3
> Good research and explanation here: https://www.scottohara.me/blog/2018/12/11/shrink-to-fit.html

This is indeed helpful. In future, I'll dig well first before making a pull request. Thank you.

PR #346, opened 2019-02-20
Update multiline config comment.

patridge:
I don't know if the existing comment reflected something I don't understand. Since the `twitter_username` field is no longer present at all, though, it felt like something that simply needed to be updated after the config file evolved.


ashmaroli:
@patridge Yes, you're right. It was missed during the changes that the config file went through.
Thanks.

PR #351, opened 2019-03-25
Allow testing with Jekyll 4.x

ashmaroli:
Test with multiple Jekyll versions on Ruby versions 2.6 and 2.4
- Test with `jekyll-4.0.0.pre.alpha1`
- Test with `jekyll-3.8.x`
- Test with any available jekyll version between `3.5.0` and `4.x`


ashmaroli:
Let's wait for `jekyll-seo-tag` to be released before merging this..

Issue #291, opened 2018-10-23
Add StackExchange / StackOverflow social links


AnthonyMastrean:
StackExchange and StackOverflow are kind of a big deal in the same community that's likely to be using Jekyll and minima. It would be _awesome_ to have Stack* username/link support!

https://stackoverflow.com/company/logos

DirtyF:
Do you want to take a stab at it and open a PR? Look at other social networks for inspiration.

AnthonyMastrean:
I can put the liquid/template stuff together, but the SVG is _way_ outside of my skill set 😝

ashawley:
The SVG for Stackoverflow already exists, so you can use that:

https://github.com/jekyll/minima/blob/6baaec2/assets/minima-social-icons.svg?short_path=fa7399f#L21

The link text is the tough bit.  See [this comment](https://github.com/jekyll/minima/pull/141#issuecomment-316476405) from the original PR:

> Since the link text is goofed up, I'm going to drop Stackoverflow unless there are any objections...
>
> ![screen shot 2017-07-19 at 2 28 13 pm](https://user-images.githubusercontent.com/358615/28383239-cbafb61c-6c8e-11e7-81cb-706f85081242.png)

All the other social sites use the username for cross-purposes of deep-linking and the human-readable name.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

ashawley:
That's a nice hack.  I'm inclined to support that.  The only other option is go the same route as Youtube in #292 and just have two fields.  Something like `stackoverflow_number` and `stackoverflow_username`?

eengstrom:
> The only other option is go the same route as Youtube in #292 and just have two fields. Something like `stackoverflow_number` and `stackoverflow_username`?

I'd be fine with that too, and there's probably some merit in picking one or the other and not mixing both.

[...]

PR #353, opened 2019-04-04
 #291 Added Stackoverflow social link

tannakartikey:
If we use the StackOveflow Story URL, there is no need to set `stackoverflow_number`. We can just use `stackoverflow_username` to render a StackOverflow Story URL.

A user can set his username by visiting https://stackoverflow.com/users/storypreferences/

Example StackOverflow Story URL: https://stackoverflow.com/story/tannakartikey


chrisfinazzo:
👋🏼 @tannakartikey, good to have you here.

In looking at this, I’d ask two things: in the opening Liquid tag, I still see “social.github” where I think you meant “social.stackoverflow”.

Second, I have an [existing][] PR where I’ve collected many of the social networks we plan to include in 4.0. You could make your request there if you wanted, but that is ultimately a group decision from @jekyll/minima, not just me.

[existing]: https://github.com/jekyll/minima/pull/333

PR #359, opened 2019-05-03
Add support for pinboard.in social links

soundasleep:
https://pinboard.in has become one option to replace the (now defunct) del.icio.us. This adds support to the minima theme for pinboard social links


DirtyF:
Hi, we don't intent to support any more social links, please override the social.html include in your project to add support to desired services.

PR #370, opened 2019-05-29
Remove google+ from social icons SVG

RDIL:
From #361


ashmaroli:
@DirtyF What happened to `@jekyllbot`?

PR #352, opened 2019-03-30
Add GitLab to social

HackingGate:
The SVG icons comes from [https://github.com/encharm/Font-Awesome-SVG-PNG](https://github.com/encharm/Font-Awesome-SVG-PNG).

- Support multiple instances

Example of _config.yml:
```yml
gitlab:
  - username: jekyll
    instance: gitlab.com
  - username: jekyll
    instance: gitlab.freedesktop.org
```

Looks:
GitLab-jekyll

Closes #325

PR #376, opened 2019-06-07
WIP: Feature/add gitlab social icon

taikedz:
This is to add Gitlab as an option to the social icons.

The only problem that leaves this still as WIP is the logo asset itself.

I took the logo from [Gitlab's press release section][gitlab-press] but it's improperly scaled (I think?) when displayed in a built Jekyll/Minima site - the icon square is an outline with empty content, I presume this is just the top left corner of the oversized icon...

I used Inkscape's "Save optimized SVG" on it, and copied over the resulting path data. I stripped what seemed to be SVG header-specific...  Beyond that however, I am unfamiliar with what I'm supposed to do to adjust the scaling / properly include a new icon.

Any page/guide I should follow I'd be happy to receive. Alternatively, would simply using a PNG be acceptable ?

Thanks

[gitlab-press]: https://about.gitlab.com/press/


taikedz:
I've just noticed the existence of #352 and #333 ; these effectively answer the questions and additions in my PR so I will close it. Thanks

PR #379, opened 2019-06-15
Fix footer stacking issue for smaller view sizes

healeycodes:
Hi 👋, this is my first PR to the minima project.

## Bug

For smaller view sizes, the footer leaves an empty column of space.

Like this:

Broken footer

## Fix

Remove the columns for these smaller view sizes by wrapping `.one-half` in a media query using `$on-large`.

So it now looks like this:

Fixed footer

The footer on larger view sizes is unchanged.

### DIFF
--- a/_sass/minima/_layout.scss
+++ b/_sass/minima/_layout.scss
@@ -292,7 +292,9 @@
 /**
  * Grid helpers
  */
-.one-half {
-  width: -webkit-calc(50% - (#{$spacing-unit} / 2));
-  width:         calc(50% - (#{$spacing-unit} / 2));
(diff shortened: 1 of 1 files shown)


alexddhuang:
A good improvement! Besides, I have observed in your picture that the last three icons don't seem to be centered. Can you improve it too?

healeycodes:
@alexddhuang thanks.

I took a look at the icon centering. The fix will likely have a few knock-on effects that I don't have the time to test for at the moment.

I'm afraid I won't be able to fix it in this PR.

DirtyF:
Thank you for the fix @healeycodes 👍

PR #382, opened 2019-06-17
Clickable post if excerpt is shown

jasontthai:
clickable_excerpt

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -23,6 +23,17 @@ layout: default
         </h3>
         {%- if site.show_excerpts -%}
           {{ post.excerpt }}
+          {% capture content_words %}
+            {{ post.content | number_of_words }}
+          {% endcapture %}
+
+          {% capture excerpt_words %}
+            {{ post.excerpt | number_of_words }}
+          {% endcapture %}
+
+          {% if content_words != excerpt_words  %}
+            <p><a href="{{site.baseurl}}{{ post.url }}">Read more</a></p>
+          {% endif %}
         {%- endif -%}
       </li>
       {%- endfor -%}
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -23,16 +23,15 @@ layout: default
         </h3>
         {%- if site.show_excerpts -%}
           {{ post.excerpt }}
-          {% capture content_words %}
-            {{ post.content | number_of_words }}
-          {% endcapture %}
+          {% assign content_words = post.content | number_of_words %}
+          {% assign excerpt_length = post.excerpt | number_of_words %}
 
           {% capture excerpt_words %}
             {{ post.excerpt | number_of_words }}
           {% endcapture %}
 
           {% if content_words != excerpt_words  %}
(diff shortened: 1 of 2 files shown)


_layouts/home.html
```diff
@@ -23,6 +23,17 @@ <h3>
         </h3>
         {%- if site.show_excerpts -%}
           {{ post.excerpt }}
+          {% capture content_words %}
```
ashmaroli: Using `{% capture %}` always results in a *string*. Wouldn't it be better if we used `{% assign .. %}` instead..?
```
{% assign content_length = post.content | number_of_words %}
{% assign excerpt_length = post.excerpt | number_of_words %}
...
```

_layouts/home.html
```diff
@@ -23,6 +23,17 @@ <h3>
...
+            {{ post.content | number_of_words }}
+          {% endcapture %}
+
+          {% capture excerpt_words %}
+            {{ post.excerpt | number_of_words }}
+          {% endcapture %}
+
+          {% if content_words != excerpt_words  %}
+            <p><a href="{{site.baseurl}}{{ post.url }}">Read more</a></p>
```
ashmaroli: Please prefer using Jekyll's `URLFilters` when URLs come into the picture:
```
<p><a href="{{ post.url  | relative_url }}">Read more</a></p>
```

_layouts/home.html
zirius: I figure content and excerpt are string, that's why I used `{% capture %}`, but I don't see why we shouldn't use `{% assign %}` instead. Let me update the PR.

_layouts/home.html
zirius: Updated the PR.

_layouts/home.html
ashmaroli: `content` and `excerpt` are strings, but the result from ` | number_of_words` isn't..

zirius: Ah right!

_layouts/home.html
```diff
@@ -23,6 +23,12 @@ <h3>
         </h3>
         {%- if site.show_excerpts -%}
           {{ post.excerpt }}
+          {% assign content_words = post.content | number_of_words %}
+          {% assign excerpt_words = post.excerpt | number_of_words %}
+
+          {% if content_words != excerpt_words  %}
+            <p><a href="{{ post.url  | relative_url }}">Read more</a></p>
```
DirtyF: This introduces an a11y issue when all read more links will have the same label. People using Voice over or similar software, will not be able to distinguish where those links get to.

zirius: I updated per your suggestion. Since the whole post is clickable, I have to strip the html from the excerpt too.

ashmaroli:
@zirius Any possible side-effects from stripping HTML from rendered excerpts?

zirius:
> @zirius Any possible side-effects from stripping HTML from rendered excerpts?

There is one side effect: html elements like links and images will be stripped and shown as regular texts.

ashmaroli:
Thank you for proposing this enhancement @jasontthai, but I feel the added complexity is a bit too much for Minima.

PR #389, opened 2019-07-13
Replace SVG with native hamburger icon

DesWurstes:
I believe this is the most elegant solution.

Before:

After:


DirtyF:
Thanks but this does not solve any problem.

PR #395, opened 2019-08-10
Supporting Table-of-Contents

alexddhuang:


ashmaroli:
So what about #365 ?

DirtyF:
Nice, but please override the theme to add such features, we want minima to stay...well minimal.

PR #397, opened 2019-08-15
Implementing a dynamic copyright feature

alexddhuang:
屏幕快照 2019-08-15 下午5 35 46


ashmaroli:
Personally, I feel this is out of Minima's scope..

DirtyF:
Thanks for your submission @alexddhuang  it's something you can do by overriding the theme. We don't want to add too much default options (they're already too much if you ask me).

PR #400, opened 2019-08-17
Add Keybase to social links

mcmd:
Added [Keybase](https://keybase.io), an encryption key directory, to the social links in the template.


DirtyF:
@jekyllbot: merge +minor

PR #403, opened 2019-08-23
Add support for LinkedIn company social media

MarkusTiede:
The URL for LinkedIn social media profiles of companies is different from the one for individuals.


DirtyF:
Thanks for your submission, please override the social.html include locally to add support for this. We don't want to support all services in this minimal theme.

PR #336, opened 2019-01-16
Allow skins to be defined and customized easily

ashmaroli:
Related to #335 

### Summary

- Rewrite the original `minima.scss` as `minima-classic.scss` default skin (codename: `"classic"`).
- All color-related Sass definitions have been extracted into `minima-classic.scss`.
- `assets/css/style.scss` simply loads a skin based on user's config settings (defaults to the "classic" skin).
- Syntax-highlighting is embedded in "the skin" since it is primarily color-related and should be adjusted harmoniously in every "skin".
- The "skin" then loads the *non-color* related styles and partials.
- The "skin" also loads a new placeholder partial that allows user to override bundled styles without affecting the upstream updates on the other partials.

PR #404, opened 2019-09-12
Solarized skins

sandervoerman:
Hi all,
I have created two skins, using the method proposed in #336. Both skins use the [Solarized](https://github.com/altercation/solarized) color scheme, one in light mode and the other in dark mode.

## minima-solarized
screenshot

[click here for minima-solarized theme preview](https://www.savoerman.nl/demo/minima-solarized/)

## minima-solarized-dark
screenshot

[click here for minima-solarized-dark theme preview](https://www.savoerman.nl/demo/minima-solarized-dark/)


ashmaroli:
Since this uses commits from PR #336, that PR needs to be reviewed and merged first..

sandervoerman:
Thanks @DirtyF, @ashmaroli!

> 
> 
> Since this uses commits from PR #336, that PR needs to be reviewed and merged first..

I understand. I am happy to update this PR should further changes be made to PR #336 (other/different variables, etc).

ashmaroli:
@sandervoerman #336 has been merged. You may update this branch.
Additionally, it'd be better if you could limit the whitespace for alignment in `minima-solarized.scss` to the current block. i.e. *The **skin color scheme values** need not align with the values in section about **Minima color scheme values***

sandervoerman:
> Additionally, it'd be better if you could limit the whitespace for alignment in `minima-solarized.scss` to the current block. i.e. _The **skin color scheme values** need not align with the values in section about **Minima color scheme values**_

Done!

ashmaroli:
@sandervoerman The header cells for a table are rendered with a *lighter* border in the `solarized-dark` skin. Is that intentional or a bug?

Screenshot_2019-11-18 This post demonstrates post content styles

sandervoerman:
> @sandervoerman The header cells for a table are rendered with a _lighter_ border in the `solarized-dark` skin. Is that intentional or a bug?

It is not intentional, but I think the bug is in _base.scss: while it sets the `table` border color to `$table-border-color`, that is overruled for `td` elements by the `td` border color, which is set to `$brand-color-light`. The latter has the same value as the former in the classic skin, but not in the solarized skins.

The commit I just pushed would fix this (for all skins) by setting the `td` border property to `$table-border-color` as well (which I am guessing was your intention when defining the `$table-border-color` variable).

After this commit, the borders all have the same color in `solarized-dark`:

Screenshot-2019-11-18

ashmaroli:
Thank you very much for working on this @sandervoerman 

@jekyllbot: merge +minor

PR #415, opened 2019-11-07
Update screenshot to reflect current rendering

ashmaroli:
So as to not confuse new users when they fork, clone and build.


ashmaroli:
Please squash and merge..

PR #424, opened 2019-12-21
Consider the font size of <code> in headers

hszhakka:
Headers may contain `<code>`s, so it's better to consider the font sizes of them too.


ashmaroli:
I think it would be better to have `<code>` inherit the font-size.
In other words, if there's a place where `code { font-size: ... }` is being declared, that needs to be changed instead.

ashmaroli:
Duplicate of #394

PR #426, opened 2020-01-03
Set the border radius of highlight

hszhakka:
I discovered the corners of code blocks look not correct currently.

屏幕快照 2020-01-03 上午11 38 08

That's because the `border-radius` of `.highlight` didn't set, so I fix it now.


ashmaroli:
Thank you for your making Minima better.
@jekyllbot: merge +fix

PR #429, opened 2020-01-07
Unguarded feed links

atcruice:
There's [a guard on the RSS link in the footer](https://github.com/jekyll/minima/blob/master/_includes/social.html#L18), yet if `site.minima.social_links.rss` is falsey, feed details still show up in the `<head>` and the home layout.

This pull request just wraps those snippets in conditions.


hszhakka:
Unnecessary.

DirtyF:
As `jeyll-feed` plugin is declared as a theme dependency, and declared in config, there's no need for those. The reason behind the guard in the footer, is that we leave it to the user if they want to display an icon or not.

Thanks for your effort though.

PR #433, opened 2020-01-16
Add a placeholder of favicons

hszhakka:
Add a placeholder of favicons, so that users do not need to override the whole `head.html` to configure favicons, instead, just override `favicons.html`.

### DIFF
--- a/_includes/favicons.html
+++ b/_includes/favicons.html
@@ -1,1 +1,1 @@
-<!-- Placeholder to allow defining custom favicons -->
+{% comment %}Placeholder to allow defining custom favicons{% endcomment %}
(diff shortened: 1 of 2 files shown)


ashmaroli:
I like the intention here. It would be better if the include contained *some markup* so that novice users don't have to search the internet to add the basic markup themselves..

Also, instead of leaving an HTML comment, I feel it'd be better (even if it would have an effect on the build time for large sites) if it was a Liquid comment (within `{% comment %}{% endcomment %}`) so that it doesn't get injected into the final HTML file.

hszhakka:
> Liquid comment

Good! I have modified it.

hszhakka:
> Nice. It would prevent 404 for missing favicons.
> Should we point people to https://realfavicongenerator.net for them to learn about favicons?

Good advice. I have added the guide steps to the comment.

ashmaroli:
Thank you @hszhakka 
@jekyllbot: merge +minor

PR #435, opened 2020-01-16
Use "em" unit in relative-font-size mixin

### LINKED ISSUE #92: code uses fixed font size
i am wondering if it might make more sense to define the CSS style for `<code>` to not use a fixed font size (it is defined as `font-size: 15px;` right now). it is legal HTML to use `<code>` in places other than just regular text (for example in headings), and defining it with a fixed font size makes these usages appear rather badly formatted.

ashmaroli:
This fixes the issue where nested elements were rendered incorrectly (e.g. `<code>` nested under headings `<h1>`..`<h6>`) because of using `px` units.

Resolves #92 
Closes #304 
Closes #372
Closes #394

### DIFF
--- a/_sass/minima/initialize.scss
+++ b/_sass/minima/initialize.scss
@@ -37,7 +37,7 @@ $on-large:         $on-laptop !default;
 }
 
 @mixin relative-font-size($ratio) {
-  font-size: $base-font-size * $ratio;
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekylbot: merge +fix

PR #436, opened 2020-01-16
Use "rem" unit in relative-font-size mixin

ashmaroli:
This fixes a miscalculation on my part in #435 

`$base-font-size * $ratio` actually translates to using `rem` units.
Using `em` units render sizes relative to the immediate parent element (which is *not* always desirable).

### DIFF
--- a/History.markdown
+++ b/History.markdown
@@ -21,7 +21,6 @@
   * Add missing rel=me attributes to social links (#386)
   * Set the border radius of highlight (#426)
   * Removed unnecessary whitespace-controls (#390)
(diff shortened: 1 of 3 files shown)


ashmaroli:
> I didn't notice any visual difference

The difference is currently visible on the home page. Try the two deploys side-by-side:
- former: https://deploy-preview-435--jekyll-minima.netlify.com/
- current: https://deploy-preview-436--jekyll-minima.netlify.com/

DirtyF:
@jekyllbot: merge +fix

PR #437, opened 2020-01-17
Assign static value to variable outside forloop

ashmaroli:
`site.minima.date_format` is a value independent of the loop context as the loop is iterating through `site.posts`. Also, since the value is static for a given build session the variable `date_format` is assigned the same value for every post in the site.


ashmaroli:
@jekyllbot: merge +bug

PR #440, opened 2020-01-17
Make the 404 page as a layout

hszhakka:
Make the 404 page as a layout, so that the user doesn't need to copy the whole HTML content into his/her repository, instead, just create a file named `404.html`, and specify its layout:

```
---
layout: 404
---
```


DirtyF:
It's needed by GitHub Pages: https://help.github.com/en/github/working-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site

hszhakka:
> Did you test it on GitHub Pages? (as long as [page is 404.html ](https://help.github.com/en/github/working-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site) it should work I guess)

Yes, I have tested on GitHub Pages. The purpose of this PR is to making users convenient to override the `404.html` page.

ashmaroli:
The `404.html` you're modifying here is for the demo site. It is not included in the theme-gem.

Moreover, `jekyll new` provides this file for new users / projects via the template within Jekyll:
https://github.com/jekyll/jekyll/blob/master/lib/site_template/404.html

So as of now a new layout for this feels a bit redundant to me.

hszhakka:
> The `404.html` you're modifying here is for the demo site. It is not included in the theme-gem.
> 
> Moreover, `jekyll new` provides this file for new users / projects via the template within Jekyll:
> https://github.com/jekyll/jekyll/blob/master/lib/site_template/404.html
> 
> So as of now a new layout for this feels a bit redundant to me.

そうか, I didn't know that the `404.html` page is from Jekyll. So I close it now.

PR #445, opened 2020-01-20
Stop prefixing CSS calc with `-webkit-`

ashmaroli:
The prefix is not needed in modern versions of webkit-based browsers:
https://caniuse.com/#feat=calc


ashmaroli:
@jekyllbot: merge +bug

PR #447, opened 2020-01-21
Design the style of task list

hszhakka:
For example:

屏幕快照 2020-01-21 下午6 00 02


DirtyF:
This feels out of sope for this minimal theme.

hszhakka:
> This feels out of scope for this minimal theme.

Task lists are widely used. Minima just needs to add a few codes to make it looks concise, I think it's not kind of very out of scope.

ashmaroli:
> Task lists are widely used

On GitHub issues and PRs, yes. But do you have any verifiable stats regarding their usage in sites using the Minima theme? There has never been a request to support / fix the styling for this..


> add a few codes to make it looks concise

Leaving the *out-of-scope* debate aside for the time-being, I think some of the proposed styling is unnecessary. You should move the task-list example in the post to below the sample showcasing numered-lists (`<ol><li/><ol/>`) and then adjust styles to match..

PR #448, opened 2020-01-22
Re-implement site footer markup and styles

ashmaroli:
## Summary of fixes
- Render a list-item (for `site.author` and `site.email`) in footer only when configured.
- Make `.social-links` a sibling of the `.footer-col-wrapper` (which now includes only `footer-col` by default). So that it is always a separate block independent of the presence of `.footer-col` divs.
- Single-line `site.description` no longer breaks the layout.
- On `$medium` viewport widths and above, use flexbox to align footer columns and social-links div. Stack vertically on smaller devices.
- Adjust social link margins so that they fold nicely on all viewport sizes when numerous icons are rendered.

### on `master`
current render

### on this branch
proposed render

## Context
Closes #446

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -4,27 +4,24 @@
   <div class="wrapper">
 
     <div class="footer-col-wrapper">
-      <div class="footer-col one-half">
-      <h2 class="footer-heading">{{ site.title | escape }}</h2>
+      <div class="footer-col">
+        <h2 class="footer-heading">{{ site.title | escape }}</h2>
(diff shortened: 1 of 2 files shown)


ashmaroli:
deploy-preview-448

/cc @hszhakka if you would like to test the social icon list alignment on mobile devices.

DirtyF:
@jekyllbot: fix +minor

DirtyF:
@jekyllbot: merge +minor

PR #449, opened 2020-01-23
Improve readability of post titles with diacritics

ashmaroli:
Translating the title `"Some articles are just so long they deserve a really long title to see if things will break well"` into Turkish yields:

## on `master`
current deploy

Note that some characters almost touch each other across lines on `master`

---

## on this branch
proposed deploy


ashmaroli:
@jekyllbot: merge +bug

PR #450, opened 2020-01-23
Support pagination through posts in home layout

### LINKED ISSUE #159: Only show a certain number of posts per page.
It seems like the index/home will continue to show all posts. What happens when a user hits 100? 1000? Even as static HTML, it is a lot to load, especially for mobile. 

Can we limit each page to like 50 or 100? Could we make it configurable?

ashmaroli:
Prior art: #150 
Resolves #159 

## Pager samples
(total of 3 pages)
First page
Second page
Last page

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -9,13 +9,21 @@ layout: default
 
   {{ content }}
 
-  {%- if site.posts.size > 0 -%}
+
+  {% if site.paginate %}
+    {% assign posts = paginator.posts %}
+  {% else %}
+    {% assign posts = site.posts %}
+  {% endif %}
+
+
+  {%- if posts.size > 0 -%}
     {%- if page.list_title -%}
       <h2 class="post-list-heading">{{ page.list_title }}</h2>
     {%- endif -%}
     <ul class="post-list">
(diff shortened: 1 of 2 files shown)


ashmaroli:
> I'd prefer browsing with "older posts" / "newer posts" links than numbers

Using just numbers have the followings pros:
- Minimal :stuck_out_tongue: 
- Reasonably same-sized pager buttons.
- *Easier* integration in localized sites. Numbers are *mostly the same* in latin-based languages.

DirtyF:
@ashmaroli as I said personal nit 😉 

@jekyllbot: merge +minor

hszhakka:
A long-awaited improvement, but don't you need to add `jekyll-pagination` in `gemspec`?

ashmaroli:
> don't you need to add `jekyll-pagination` in `gemspec`?

No. Pagination is an opt-in feature.
The config setting `paginate` is needed for `jekyll-paginate` plugin (supported on GitHub Pages). I don't know if `jekyll-pagination` accepts the same config.

Users wishing to add pagination should do the needful at their end. Minima 3.0 just *supports* using it out-of-the box.

PR #455, opened 2020-02-05
Fix a bug of incorrect parsing of page.modified_date

hszhakka:
It seems that `page.modified_date` didn't consider the timezone of the site, so we have to add an additional filter `date_to_xmlschema`.


_layouts/post.html
```diff
@@ -13,7 +13,7 @@ <h1 class="post-title p-name" itemprop="name headline">{{ page.title | escape }}
       {%- if page.modified_date -%}
         ~ 
         <time class="dt-modified" datetime="{{ page.modified_date | date_to_xmlschema }}" itemprop="dateModified">
-          {{ page.modified_date | date: date_format }}
+          {{ page.modified_date | date_to_xmlschema | date: date_format }}
```
ashmaroli: `{{ page.modified_date | date_to_xmlschema }}` will now get computed twice.
So let us stash the result in a variable and then render the result:
```html
{% assign mdate = page.modified_date | date_to_xmlschema %}
<time class="dt-modified" datetime="{{ mdate }}" itemprop="dateModified">
  {{ mdate | date: date_format }}
</time>
```

hszhakka: Good

DirtyF:
@jekyllbot: merge +fix

PR #456, opened 2020-02-11
Markdownify title

hszhakka:


ashmaroli:
Sorry, this enhancement is outside the scope of Minima.

The other reason being that this will lead to *unnecessary* string duplication (and therefore memory allocation) for non-markdown titles

hszhakka:
> Sorry, this enhancement is outside the scope of Minima.
> 
> The other reason being that this will lead to _unnecessary_ string duplication (and therefore memory allocation) for non-markdown titles

Outside the scope of Minima? Titles with some Markdown formats are frequently used, especially for programmers, titles with code are normal so that they need `markdownify`.

Since titles are normally short -- comparing with text, the string duplication maybe not so horrible.

ashmaroli:
> Titles with some Markdown formats are frequently used..

Could you post some links to repos that uses Markdown formatted titles in the front matter and is rendered with Minima theme.

> Since titles are normally short

No arguments here. Titles are *generally* shorter.

But my disagreement is not with the size of allocated strings but the number. Each filter (in `markdownify | remove: "<p>" | remove: "</p>"`) duplicates the given input string. So if I have `"Hello World"` as the title for a post, it'll get duplicated once by `markdownify` filter and twice again by the `remove` filter. These extra allocations are technically *garbage* since they don't serve a purpose.
Now, imagine having these cruft generated for *every post, document and page* in the site (because they are just plain-text)..

Why should a major size of users have to be subject to *garbage* so that a minority of users can  have styled titles..? A minimal theme should also be minimal (within reasonable limits) in resource usage as well.

---

If you can provide adequate evidence that this is going to have a positive effect for a major size of users, I'll reconsider.

DirtyF:
Let keep things simple for now, minima is used by thousands of users and this hasn't surfaced. Also, themes are easily overwritable, minima don't aim at solving all possible use cases.

PR #458, opened 2020-02-12
Set margin-bottom of pagination

hszhakka:


DirtyF:
@jekyllbot: merge +minor

PR #462, opened 2020-02-13
Export a variable to let users to custom the font family of code

hszhakka:
Before:

Screenshot 2020-02-13 at 9 39 06 PM

After:

Screenshot 2020-02-13 at 9 39 49 PM

### DIFF
--- a/_sass/minima/_base.scss
+++ b/_sass/minima/_base.scss
@@ -153,6 +153,7 @@ blockquote {
  */
 pre,
 code {
+  font-family: $code-font-family;
   font-size: 0.9375em;
   border: 1px solid $brand-color-light;
   border-radius: 3px;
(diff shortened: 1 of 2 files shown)


ashmaroli:
Couple of stuff:
- Even though `"Monospace"` is a font, you should include `monospace` as the last fall back.
- Since the base-font has lots of choices, I think the code font should have more choices too. I think you should add `"Consolas", "Roboto Mono", "Ubuntu Mono" to the list.
- "Courier New" is like `"Times New Roman"` for text. It should have lower priority than other fonts.
- As always for such PRs, please post a *before* and *after* set.

hszhakka:
> Couple of stuff:
> 
> * Even though `"Monospace"` is a font, you should include `monospace` as the last fall back.
> * Since the base-font has lots of choices, I think the code font should have more choices too. I think you should add `"Consolas", "Roboto Mono", "Ubuntu Mono" to the list.
> * "Courier New" is like `"Times New Roman"` for text. It should have lower priority than other fonts.
> * As always for such PRs, please post a _before_ and _after_ set.

OK, I have added some more fonts.

DirtyF:
@jekyllbot: merge +minor

PR #473, opened 2020-02-19
Add a placeholder to allow defining custom footer

hszhakka:


ashmaroli:
I was fine with the `custom-head` *hook* because it had value for a greater number of users &mdash; easily add javascript / external stylesheets, fonts / favicon metadata ... etc.

But the footer is a part of the visible aspect of the site. The placement of the *hook* therefore cannot be standardized. What if someone wanted the copyright placed before the social-media list? Or perhaps as the first thing on the footer before other items currently in the footer.

The only broad-spectrum use for this could be to inject scripts before the closing `</body>` tag..
So, this is being withheld for the time being.

PR #477, opened 2020-02-20
Show the words of post

Tianao:


hszhakka:
Out of the scope. Besides, `{{ content.size }}` doesn't calculate the number of words, instead, it calculates the number of characters.

ashmaroli:
Sorry @Tianao, Minima is intended to be *a minimal theme* by default.

PR #478, opened 2020-02-21
Improve customizability of the theme

### LINKED ISSUE #474: Cannot override custom-variables.scss and custom-styles.scss
## Summary

I am using jekyll-remote-theme to import the `master` branch of Minima, but I discover that I cannot override `custom-variables.scss` and `custom-styles.scss`. I put these two files into `_sass/minima/` in my repository, but after running `bundle exec jekyll serve`, the real imported files are still from the Minima repo.

Here is a CSS map:

Screenshot 2020-02-20 at 6 47 04 PM

Can anyone help me?

## Environment

- Jekyll: `4.0.0`
- Bundler: `2.1.4`
- jekyll-remote-theme: `0.4.1`
- Minima: the `master` branch

ashmaroli:
## Summary

Change import paths in skins so that the partials are not so tightly coupled.
Also updates the README to add clarity on the theme contents and their overriding at source directory.

## Context

Resolves #474

### DIFF
--- a/assets/css/style.scss
+++ b/assets/css/style.scss
@@ -2,4 +2,6 @@
 # Only the main Sass file needs front matter (the dashes are enough)
 ---
 
-@import "minima-{{ site.minima.skin | default: 'classic' }}";
+@import
+  "minima-{{ site.minima.skin | default: 'classic' }}",
+  "minima/initialize";
--- a/README.md
+++ b/README.md
@@ -88,9 +88,11 @@ Refer the [skins](#skins) section for more details.
 ### Assets
 
 Refers to various asset files within the `assets` directory.
-Contains the `css/style.scss` that imports sass files from within the `_sass` directory. This `css/style.scss` is what gets processed into the theme's main stylesheet `main.css` called by `_layouts/default.html` via `_includes/head.html`.
 
-This directory can include sub-directories to manage assets of similar type (`img`, `fonts`, `svg`), and will be copied over as is, to the final transformed site directory.
+  - `assets/css/style.scss` &mdash; Imports sass files from within the `_sass` directory and gets processed into the theme's
+    stylesheet: `assets/css/styles.css`.
+  - `assets/minima-social-icons.svg` &mdash; A composite SVG file comprised of *symbols* related to various social-media icons.
+    This file is used as-is without any processing. Refer [section on social networks](#social-networks) for its usage.
 
 
 ### Plugins
--- a/README.md
+++ b/README.md
@@ -76,10 +76,11 @@ Refers to `.scss` files within the `_sass` directory that define the theme's sty
   - `minima-classic.scss` &mdash; The core file imported by preprocessed `css/style.scss`, it defines the variable defaults for
     the "classic" skin of the theme.
   - `minima/initialize.scss` &mdash; A component that defines the theme's *skin-agnostic* variable defaults and sass partials.
-  - `minima/custom-variables.scss` &mdash; A hook that allows overriding variable defaults and mixins. (*Note: Cannot override styles*)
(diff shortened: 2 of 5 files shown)


ashmaroli:
@hszhakka I've updated the import logic and info in the README. Interested in giving this a test-drive?

assets/css/style.scss
```diff
@@ -2,4 +2,6 @@
 # Only the main Sass file needs front matter (the dashes are enough)
 ---
 
-@import "minima-{{ site.minima.skin | default: 'classic' }}";
+@import
+  "minima-{{ site.minima.skin | default: 'classic' }}",
```
hszhakka: Since `minima-classic.scss`, `minima-solarized.scss`, and `minima-solarized-dark.scss` are only color-related, I think put these files into `_sass/minima/skins/` would make their purpose clear. The file structure looks like below:

```
_sass/
  minima/
    skins/
      classic.scss
      solarized.scss
      solarized-dark.scss
    _base.scss
    _layout.scss
    initialize.scss
```

Then this line is modified to 

```scss
@import
  "minima/skins/{{ site.minima.skin | default: 'classic' }}",
```

hszhakka:
> @hszhakka I've updated the import logic and info in the README. Interested in giving this a test-drive?

I have tested it. It's OK. Now just copying `custom-variable.scss` and `custom-styles.scss` can do overriding, but I have a little [proposal](https://github.com/jekyll/minima/pull/478#discussion_r382436193) about the file structure, hope you can check it.

assets/css/style.scss
ashmaroli: This is a good suggestion. Therefore, I'll let you take full credit for it.
Feel free to submit a PR with the change and associated updates to the README after this PR gets merged.

hszhakka: OK

ashmaroli:
> I feel we're over-engineering a minimal theme here.

@DirtyF I think you're missing the point of this PR. Currently with `master`, to use the `custom-styles.scss` and `custom-variables.scss` hooks, they've to go through confusing hoops. This PR remedies that to the simplicity intended by the hooks.

DirtyF:
I'm just saying these docs are way too complicated for a minimal theme.
We keep adding features.
just wanna reinforce on the theme scope: just show me how to use the theme in the configuration file with basic options.
Advanced usage is you ask me it out of scope for this theme.

JJC1138:
This is working great for me too. Thank you again.

I noticed that `custom-variables.scss` is imported after the skin. A consequence of that is that if I customize `$brand_color` then I'd have to also customize the other things that the skin defines that are derived from that.

Do you think it would be better to include `custom-variables.scss` before the skin to avoid that issue? I'd be happy to do a PR with that if so.

ashmaroli:
@JJC1138 Thanks for the feedback. I designed it this way because the variables defined in a *skin* are to be limited to the skin itself.
Since you're customizing `$brand_color`, it would translate into altering the entire skin. You're no longer depending on the *predefined skin*.

In that event, all you need to do is save the skin with a new name and use that. For example, say you want `$brand_color` to be *crimson red*, then you may do the following:
- Copy `_sass/minima/skins/classic.scss` (if that is preferable as the starting point) to your source directory
**as `_sass/minima/skins/classic-crimson.scss`**.
- Change `$brand_color` to the desired value in `<source>/_sass/minima/skins/classic-crimson.scss`
- Edit config file to load the custom skin:
  ```yaml
  # _config.yml

  minima:
    skin: "classic-crimson"
  ```

The above is the intended workflow with Minima 3.0. If there's a *bug* in the above workflow (i.e., doesn't work as expected), I'd be happy to review a PR resolving it.

JJC1138:
Ah, I see. Skins aren't intended to partially overridden in the way that I was proposing, and it makes more sense to copy the whole skin to make changes like that. Thank you!

PR #480, opened 2020-02-23
[chore] Normalize Travis CI builds

ashmaroli:
- Installing via a shell script was messing with the build cache.
- Jekyll ~> 3.5 is the same as Jekyll ~> 3.8.0.
- Test with just one Ruby version since there isn't any actual Ruby code bundled with the gem (except for the gemspec) that warrants testing with multiple Ruby versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,13 @@ rvm:
   - 2.6
   - 2.4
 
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyllbot: merge +dev

PR #482, opened 2020-02-26
Add placeholders for customizing post header and footer

hszhakka:


ashmaroli:
@hszhakka Thanks but to be clear, Minima does not intend to incorporate any more placeholders than already available.

PR #483, opened 2020-02-26
Fix a bug of relative-font-size

hszhakka:
Using `rem` in `relative-font-size` may cause the layout abnormal when adjust the base font size. For example, after setting `$base-font-size` to `30px`, the homepage looks like:

FireShot Capture 009 - Your awesome title - Write an awesome description for your new site h_ - 127 0 0 1

After replacing `rem` with multiplication with `$base-font-size`, it looks correct:

FireShot Capture 010 - Your awesome title - Write an awesome description for your new site h_ - 127 0 0 1

### DIFF
--- a/_sass/minima/initialize.scss
+++ b/_sass/minima/initialize.scss
@@ -38,7 +38,7 @@ $on-large:         $on-laptop !default;
 }
 
 @mixin relative-font-size($ratio) {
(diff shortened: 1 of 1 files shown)


ashmaroli:
This was changed to avoid `px` units.
Instead I'd prefer that you change all elements that change size abnormally.

PR #484, opened 2020-02-26
Use multiplication with base-font-size to replace rem on some elements

hszhakka:
Use multiplication with base-font-size to replace rem on some elements to avoid abnormal size when adjusting base-font-size.

Relate: #483


hszhakka:
Not good. Sorry.

PR #485, opened 2020-02-26
Specify the font size of the root element

hszhakka:
Since `relative-font-size` uses `rem`, for avoiding the abnormal size cause by adjusting `base-font-size`, we need to set the font size of the root element to `base-font-size`.


DirtyF:
@jekyllbot: merge +fix

PR #486, opened 2020-02-27
Set only the border of the block code

hszhakka:
Eliminating the border of the inline code can make it more fused with the normal text without being obtrusive.

Before:

Screenshot 2020-02-27 at 11 22 42 AM

After:

Screenshot 2020-02-27 at 11 22 11 AM


MichaelCurrin:
The screenshots look identical to me. Perhaps you can focus a pair of screenshots on a narrower border area you are referring to.


Also conflicts need to be fixed, so rebase onto master.

PR #487, opened 2020-02-27
Set hyphens to auto to improve the alignment of text

hszhakka:
Before:

Screenshot 2020-02-28 at 1 19 15 AM

After:

Screenshot 2020-02-28 at 1 19 23 AM

### DIFF
--- a/_sass/minima/_base.scss
+++ b/_sass/minima/_base.scss
@@ -34,6 +34,11 @@ body {
   overflow-wrap: break-word;
 }
 
+p {
+  -webkit-hyphens: auto;
+      -ms-hyphens: auto;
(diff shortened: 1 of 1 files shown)


PanderMusubi:
This was also a default in e.g. WordPress sites. However, some languages have outdated hyphenation patterns and it caused for those languages more bad then good. Many websites have it turned off by default because of this. Personally, I like hyphenation but as a default setting not.

mattr-:
Thanks for contributing!

We'll stick with word wrapping rather than hyphens here. I think hyphenation is a addition folks can make to their site's stylesheet if that's what they want but I don't think it's a good default.

PR #490, opened 2020-02-29
Use relative border-radius to make elements consistent after scaling

hszhakka:


mattr-:
This looks too much like personal customizations so I'm closing it. If a0e96bb is the relevant commit, then let's get that to be the only thing in the PR.

PR #493, opened 2020-03-01
Allow users to select the comments provider

hszhakka:
The users can select their comments provider by configuring:

```yml
minima:
  comments_provider: # Configure the comments provider here (default: disqus)
```

For example, if the comments provider is `discourse`, the user needs to create a file `_includes/discourse_comments.html` and put all script code of `discourse` into that file.

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -30,9 +30,10 @@ layout: default
     {{ content }}
   </div>
 
-  {%- if site.disqus.shortname -%}
-    {%- include disqus_comments.html -%}
-  {%- endif -%}
+  {% if page.comments != false %}
+    {% capture comments_provider %}{{ site.minima.comments_provider | default: "disqus" }}{% endcapture %}
+    {% include {{ comments_provider }}_comments.html %}
+  {% endif %}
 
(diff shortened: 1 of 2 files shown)


DirtyF:
What  comment providers can we list here? Do they all implement the same as Disqus?

_layouts/post.html
```diff
@@ -30,9 +30,10 @@ <h1 class="post-title p-name" itemprop="name headline">{{ page.title | escape }}
     {{ content }}
   </div>
 
-  {%- if site.disqus.shortname -%}
-    {%- include disqus_comments.html -%}
-  {%- endif -%}
+  {% if page.comments != false and jekyll.environment == "production" %}
```
MichaelCurrin, line 33: If you get the value and it is not set on the page, it will be null and then `null != false` will be truthy.

```
---
---
{{ page.fizz | jsonify }}
```
Result is `null`

So while `page.comments != false` is unreliable, then `page.comments == true` will only be truthy if the value is true and not if it is false or missing.

```suggestion
  {% if page.comments and page.comments == true and jekyll.environment == "production" %}
```

OR

```suggestion
  {% if page.comments and page.comments == true and jekyll.environment == "production" %}
```

PR #496, opened 2020-03-18
remove comparison for boolean

junhan-z:
page.comments is enough

### DIFF
--- a/_includes/disqus_comments.html
+++ b/_includes/disqus_comments.html
@@ -1,4 +1,4 @@
-{%- if page.comments != false and jekyll.environment == "production" -%}
+{%- if page.comments and jekyll.environment == "production" -%}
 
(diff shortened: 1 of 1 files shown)


ashmaroli:
@junhan-z Actually there is a small detail you're not seeing though.
Currently, the comments are rendered by default. i.e. even if you don't define a `comments` key in the front matter. One can *disable* the comments for a particular page by explicitly having `comments: false` in the front matter.

With your change, the behavior changes.
The comments would **not be rendered** by default. One has to *explicitly* have `comments: true` in all pages and documents to render comments.

junhan-z:
@ashmaroli 
I see, I didn't consider that, now it makes sense to me, thanks!

junhan-z:
Maybe we should add some comments near it?

PR #498, opened 2020-03-28
Add support for Google Scholar profile

frgossen:


ashmaroli:
@frgossen I regret to inform you that we won't be bundling additional social icons with Minima theme. Sorry

PR #452, opened 2020-01-26
Extract more color rules into skins

ashmaroli:
- Allow border-colors to be altered easily per skin.
- Allow link:hover state and site-title-color to be altered per skin.
- Replace *use* of `$brand-color-light` and `$brand-color-dark` with tint-agnostic-variables (since *dark* skins use lighter values for contrast).

PR #335, opened 2019-01-15
Introduce "Dark Minima" skin

xxyzz:
Resolves #143 

Color palette based on https://primer.style/ (ref: https://github.com/jekyll/minima/issues/143#issuecomment-435308808)

**update:** jekyll-minima-dark-skin

PR #506, opened 2020-04-08
Add skin classic-dark

jeremyhalin:
Hi,
Here is my proposal for a `classic-dark` skin.
What do you think?

Preview classic-dark skin

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -12,6 +12,7 @@ show_excerpts: false # set to true to show excerpts on the homepage
 # Minima date format
 # refer to https://shopify.github.io/liquid/filters/date/ if you want to customize this
 minima:
+  skin: classic-dark
   date_format: "%b %-d, %Y"
 
   # generate social links in footer
--- a/_sass/minima/skins/classic-dark.scss
+++ b/_sass/minima/skins/classic-dark.scss
@@ -1,0 +1,95 @@
+@charset "utf-8";
+
+// Classic Dark skin
+// ==============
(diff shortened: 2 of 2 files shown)


ashmaroli:
Thank you very much @jeremyhalin. If you don't mind, I'm going to leave some comments below. You're free to consider / reject them..

- I think the name `classic-dark` is verbose. How about just `dark`?
- I personally feel the current background is too black. How about `#181818` to reduce the contrast?
- I see that borders for the social-icons are now `transparent`. Why?
- Once #452 gets merged, I think you'll have more control over the tweaking.. Will you be able to revisit this once that gets merged..?

jeremyhalin:
- Sure `dark` is a better name.
- For the black background, I used `#121212` from [Dark theme - Material Design](https://material.io/design/color/dark-theme.html#properties)
- I made borders transparent to hide it honestly, personal taste I guess so maybe I could set a white color with less opacity
- sure thing, when #452 will get merged?

jeremyhalin:
I see there is already an ongoing work about a dark theme #335

ashmaroli:
> set a white color with less opacity

I'd prefer setting a solid grey that matches the resulting color instead of a white with an alpha channel.


> when #452 will get merged?

Can't really say.. It is up for review from one of the other maintainers.

ashmaroli:
> already an ongoing work about a dark theme

Yes. But it hasn't been revisited.. So can't really say if it'll come up to speed..

DirtyF:
Duplicate of #335

PR #507, opened 2020-04-12
fix: default to having no social_links

ticoombs:
## Issue:
Having default social links under the theme configuration file, results in those items showing up, if the user does not override them with their own. 

```
_config.yml
 minima:
   social_links:
     mastodon:
       - username: test
         instance: example.com
```

## Result:

social-icons-default

## Solution:
Removal of default theme social configuration settings to ensure the only social_links are ones the user defines.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -16,8 +16,8 @@ minima:
 
   # generate social links in footer
   social_links:
-    twitter: jekyllrb
-    github:  jekyll
+    # twitter: jekyllrb
+    # github:  jekyll
     # devto: jekyll
     # dribbble: jekyll
     # facebook: jekyll


ashmaroli:
@ticoombs The change you're proposing should not be necessary.
- If you're using this theme by forking the repository, then you must make the desired change in that fork.
- Or if you're using this theme as a **remote_theme** along with **Jekyll 4**, then it is a bug with Jekyll 4. Try using Jekyll 3.8.6 in your Gemfile. If the issue gets resolved when you downgrade, please  open an issue ticket at the `jekyll/jekyll` repository.

Thank you.

ticoombs:
hey @ashmaroli 

Looks like its a bug with **Jekyll 4**. Sorry for the wrong repo. 

Cheers!

PR #514, opened 2020-04-29
Add tumblr social link

laxa88:
Preview:

minima-tumblr

Tumblr logo SVG taken from: https://www.svgrepo.com/svg/141732/tumblr-logo

Note: I resized the SVG to 60% to match the size of other social icons, using online tool: https://aydos.com/svgedit/


MichaelCurrin (approved):

MichaelCurrin:
Needs update from master but looks good

ashmaroli:

laxa88:
@ashmaroli Just curious, I don't see any harm in adding this (and other missing social links); is there a reason why?

ashmaroli:
Because Minima was originally intended to be a minimal *get-started-with-jekyll* theme that provides the bare essentials to help novice Jekyll users get onboard with the program and focus on writing content. *Minimal not only in the aesthetics but also in the features*.
Somewhere along the dev road, the above intention got diluted to result in a minimal theme that would encourage users to customize the theme, and in the process, get acquainted with managing a theme with Jekyll.

Social media platforms are many in number. We too thought there's isn't harm in adding support for some more but then soon realized that supporting each and every popular platform is going to get unwieldy and go against the *minimal* theme idea.

PR #535, opened 2020-08-20
Conditionally load GFM-parser gem in CI builds

ashmaroli:
So that CI can reliably run with Jekyll 3.8, 3.9 and 4.x


DirtyF (approved):

ashmaroli:
@jekyllbot: merge +dev

PR #536, opened 2020-08-25
Fix for disqus_comments.html

lukas2:
Disqus would not load on my blog until I have removed these lines.

### DIFF
--- a/_includes/disqus_comments.html
+++ b/_includes/disqus_comments.html
@@ -2,11 +2,6 @@
 
   <div id="disqus_thread"></div>
   <script>
-    var disqus_config = function () {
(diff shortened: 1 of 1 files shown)


_includes/disqus_comments.html
```diff
@@ -2,11 +2,6 @@
 
   <div id="disqus_thread"></div>
   <script>
-    var disqus_config = function () {
-      this.page.url = '{{ page.url | absolute_url }}';
-      this.page.identifier = '{{ page.url | absolute_url }}';
-    };
-
```
ngzhio, line 9: These lines are from official Disqus. I think that `page.url` may have some errors. Maybe you should check what errors reported in the console.

PR #550, opened 2020-10-29
add patreon to social

chrisdecker1201:
I really would love to add patreon as a social link.

The patreon icon I copied from here:
https://www.svgrepo.com/svg/306541/patreon


ashmaroli:

Issue #517, opened 2020-05-03
Trying to use custom css on Minima Theme Jekyll


uxdiva:
I've just installed Jekyll and I'm trying to customize the base theme Minima but it won't seem to grab my custom styles.

Reading the documentation... it says that I should create a file `/assets/css/style.scss` -- but the system ignores that file.

There is an option to extend the head section by creating a file `/_layout/custom-head.html` . I tried to do that and just create my own css file and point to it. But it doesn't look like Jekyll is grabbing my custom file either. 

Please help. I've been trying to figure this out all day.

ashmaroli:
Hello @uxdiva,
Sorry to hear that you're facing issues with customizing the theme.

Since you are a Jekyll novice, I'm going to assume that you just ran `jekyll new <something>` and started to customize the theme based on the Jekyll documentation. If the assumption is correct, then you're using `v2.5.1` of Minima gem and therefore, the correct theme documentation for that version is at
[this link](https://github.com/jekyll/minima/tree/v2.5.1).

However, if my assumption is wrong and that you're using Minima via a different method, do provide the details on how exactly.

uxdiva:
Yes -- I just installed using Jekyll new. I will check the documentation for the link above and let you know if I have any more questions. THANKS :D!

patrickhlauke:
Crikey, I've been trying for days to get some custom styling for GitHub pages that use the built-in minima theme, despairing at how the instructions on https://github.com/jekyll/minima#customizing-templates and https://help.github.com/en/github/working-with-github-pages/adding-a-theme-to-your-github-pages-site-using-jekyll had no effect. Luckily stumbled across this issue here, and indeed ... https://github.com/jekyll/minima/tree/v2.5.1#customization did the trick.

Wonder if it's worth specifying...somewhere (assuming some people will land on this repo in the hope, like me, to be able to do gentle modifications to their gh-pages setup)

raxod502:
I assume that `custom-head.html` is a new feature. Would it be possible to get a release cut? Copying the whole header feels pretty silly, but it seems that's the only way to add anything using the latest release 2.5.1?

gstotts:
@patrickhlauke I've been having the same problem but after following https://github.com/jekyll/minima/tree/v2.5.1#customization and creating an `/assets/main.scss` in my repo, still am not seeing what I expect.  Did that work for you or was there more you needed?

jekyllbot:
[...]

PR #564, opened 2020-12-17
FIX: Edit minima/_includes/head.html to include assets/css/style.scss…

njsch:
FIX: Edited minima/_includes/head.html to include assets/css/style.scss instead of assets/css/style.css, so that the file is actually found when needing to be included.  Sticking with file extension convention for other style sheets within the minima directory.  May or may not fix one or two currently open issues but not sure.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -1,9 +1,9 @@
 <head>
-  <meta charset="utf-8">
+]  <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


njsch:
Is this related to issue #517? FYI @uxdiva.

ashmaroli:
@njsch One should always use the `<link/>` tag to link to the processed output file instead of a preprocessed file.

PR #567, opened 2020-12-24
add feature

TakreemAkhter:
reading time indicator for posts

### DIFF
--- a/_includes/reading-time.html
+++ b/_includes/reading-time.html
@@ -1,0 +1,6 @@
+{% capture words %}
+{{ content | number_of_words | minus: 180 }}
+{% endcapture %}
+{% unless words contains '-' %}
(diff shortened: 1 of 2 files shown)


sziv2p:
1. You didn't provide the `reading-time.html` file.
2. This feature might be out of the scope of this theme.

TakreemAkhter:
I agree with your second point but it is a good addition and promotes user engagement.

sziv2p:
> I agree with your second point but it is a good addition and promotes user engagement.

It depends on what the maintainers @ashmaroli @DirtyF think. I have also submitted some small improvements before, but they have all been rejected because they are beyond the scope of this theme.

TakreemAkhter:
Thanks

PR #585, opened 2021-04-16
[2.5-stable] Give simple color override instructions

riking:
Additionally, encourage the use of the forward-compatible `minima/custom-styles` name for any changes beyond SCSS variables.

See also https://github.com/jekyll/jekyll/pull/8642, which will link to the 2.5-stable branch until the new version is promoted to stable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,19 +94,22 @@ The title for this section is `Posts` by default and rendered with an `<h2>` tag
 To override the default structure and style of minima, simply create the concerned directory at the root of your site, copy the file you wish to customize to that directory, and then edit the file.
 e.g., to override the [`_includes/head.html `](_includes/head.html) file to specify a custom style path, create an `_includes` directory, copy `_includes/head.html` from minima gem folder to `<yoursite>/_includes` and start editing that file.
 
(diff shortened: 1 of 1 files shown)


riking:
> cannot load such file -- kramdown-parser-gfm

Looks like solidly not my fault.

README.md
```diff
@@ -94,19 +94,22 @@ The title for this section is `Posts` by default and rendered with an `<h2>` tag
...
+```scss
+---
+# empty front matter
+---
+
+// Override any SCSS variables here.
+// $background-color: #fdfdfd;
+// $brand-color: #2a7ae2;
+@import "minima";
+// Need more than just color changes? Fill out the _sass/minima/custom-styles.scss file.
```
mattr-, line 108: Would you mind updating the comment here? I can't tell if the user needs to add `_sass/minima/custom-styles.css` to their site? Import it as in the example below? There's some text a couple of lines below that looks like it gives some additional instructions. Perhaps we should combine that one with this comment here?

riking: Yes, you add it as a new file to your site _in addition_ to this.

light-and-ray:
What is about this change?

light-and-ray:
I had big problem with brand-color override using GitHub Pages, and could do it only after reading readme from this abandoned PL 💔

PR #591, opened 2021-06-23
Remove files generated by OS

nickmccurdy:
OS files are unrelated to the project, and therefore should be in a global gitignore.


mattr-:
Keeping these here improves the developer experience for those folks who don't have a global gitignore file. Declining to merge this PR.

PR #601, opened 2021-08-29
Enhance theme to allow switching skins client-side

ashmaroli:
## Key points

- Generate a complete CSS file for each skin:
  - `assets/css/style.css` (default stylesheet / *classic* skin)
  - `assets/css/dark.css` (*dark* skin)
  - `assets/css/solarized.css` (*solarized* skin)
  - `assets/css/solarized-dark.css` (*solarized-dark* skin)
- Use javascript to render a `<select/>` element in the site-footer **to allow switching stylesheets on a deployed site**.
- Introduce `custom-scripts.html` placeholder to allow adding more scripts before the closing `</body>` tag.

---

Live Demo: https://ashmaroli.github.io/minima/

### DIFF
--- a/_includes/custom-scripts.html
+++ b/_includes/custom-scripts.html
@@ -1,0 +1,3 @@
+{% comment %}
+  Placeholder to allow inserting <script/>s at the end of the page body.
+{% endcomment %}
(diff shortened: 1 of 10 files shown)


mattr-:
I like this idea. What do you think about moving the skin switching bits to a documentation example? This will give folks an example that they can use to seamlessly implement something for their own site and allows us to avoid implementing color scheme switching functionality directly to the theme.

PR #607, opened 2021-10-03
Added system-ui as a cross-platform default font

vedranmiletic:
This is what happens on Linux when Segoe UI Emoji is available, but Segoe UI isn't:

minima-before

`system-ui` is provided by modern browsers as [a cross-platform default font](https://caniuse.com/font-family-system-ui). This is how it looks with `system-ui` in the font list:

minima-after

It was a controversial choice back in 2017. because it did not address i18n well and therefore it was removed from Bootstrap twbs/bootstrap#22377. However, it was added back in v5 twbs/bootstrap#30561.

### DIFF
--- a/_sass/minima/initialize.scss
+++ b/_sass/minima/initialize.scss
@@ -2,7 +2,7 @@
 
 // Define defaults for each variable.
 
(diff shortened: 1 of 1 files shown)


mattr- (approved):

mattr-:
@jekyllbot: merge +minor

mattr-:
Thanks for the contribution!

vedranmiletic:
Thank you for merging the change.

PR #608, opened 2021-10-03
Add a basic workflow file for building minima

mattr-:
This will replace the travis CI configuration

PR #609, opened 2021-10-04
Improve CI workflow configuration

ashmaroli:
Follow-up to #608

- Trigger workflow for `pull_request` event as well.
- Improve job name in statuses.
- Set env once for all steps.
- Run `bundle install` as part of `setup-ruby` action.


ashmaroli:
@jekyllbot: merge +dev

PR #621, opened 2021-12-21
Translation

Cathe313:
French translation of site's main pages


Cathe313:
Accidentally pulled to the wrong repo

PR #650, opened 2022-04-27
Add automatic Mastodon verification

ironiridis:
Mastodon supports [link verification](https://docs.joinmastodon.org/user/profile/#verification) by checking a page to see if it has a link back to the canonical URL of the Mastodon profile. This modification adds the `rel="me"` attribute to any Mastodon profile links to support this verification.


ashmaroli:
Thank you for the proposal but the 2.5-stable branch is not accepting external contributions.

PR #662, opened 2022-06-27
Add minima 3 installation to readme

### LINKED ISSUE #661: Installation instructions for minima 3
Right now the README documentation is very weird because although most people are probably using Minima 2, the documentation explains mostly how to use Minima 3, but it also only shows how to install the Minima 2 theme using the gem 🤔. It would be nice to have installation instructions for both Minima 2 and 3, an explanation between the (small, albeit relevant) differences between both differences including the distribution method (Minima 3 is only available through GitHub AFAIK).

### TLDR

Add to the README this line:

```yaml
gem "minima", github: "jekyll/minima"
```

I'll fork minima this week and create a PR for these changes if no one does it before that.

DoodlesEpic:
Closes #661

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,11 @@
 # minima
 
-*Minima is a one-size-fits-all Jekyll theme for writers*. It's Jekyll's default (and first) theme. It's what you get when you run `jekyll new`.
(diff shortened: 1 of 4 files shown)


DoodlesEpic:
Wrong branch, sorry. Will create another one with the right branch.

PR #665, opened 2022-06-27
Remove default twitter and github links

### LINKED ISSUE #664: Can't remove Twitter nor GitHub links
It looks like the default _config.yml has uncommented values for the Twitter and GitHub, which means that websites using Minima cannot remove Twitter and GitHub socials if they so wish.

DoodlesEpic:
This will not please the demo website but now it means that anyone using Minima gets to disable Twitter and GitHub social links.
Closes #664

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -16,8 +16,8 @@ minima:
 
   # generate social links in footer
   social_links:
-    twitter: jekyllrb
-    github:  jekyll
+    # twitter: jekyllrb
+    # github:  jekyll
     # devto: jekyll
     # dribbble: jekyll
     # facebook: jekyll


getaaron:
we should do the same with

```
author:
  name: GitHub User
  email: your-email@domain.com
```

since there's no way to _not_ set those now

shevoann (approved):

_config.yml
```diff
@@ -16,8 +16,8 @@ minima:
 
   # generate social links in footer
   social_links:
-    twitter: jekyllrb
-    github:  jekyll
+    # twitter: jekyllrb
```
shevoann, line 19: _config.yml

_config.yml
```diff
@@ -16,8 +16,8 @@ minima:
...
+    # github:  jekyll
```
shevoann, line 20: f

PR #673, opened 2022-09-03
Add google scholar.

juicedatom:
This is an example PR which adds the google scholar icon into the list of possible social links.


juicedatom:
Here's a quick image of me uncommenting the google scholar embedded links. I wanted to give some random, but valid link for the scholar link so I just made it Albert Einstein.

scholar_001

_sass/minima/_base.scss
```diff
@@ -241,6 +241,9 @@ pre {
   width: 16px;
   height: 16px;
   display: inline-block;
+  /* This is needed just for google scholar. */
+  filter: grayscale(100%);
```
juicedatom, line 245: For some reason google scholar needed this to stay gray like all the other links. Were the other links just gray by default? I don't really know web dev so it was hard for me to trace where currentColor was coming from.

_includes/social.html
```diff
@@ -18,4 +18,5 @@
...
   {%- if social.microdotblog -%}<li><a rel="me" href="https://micro.blog/{{ social.microdotblog | cgi_escape | escape }}" target="_blank" title="{{ social.microdotblog | escape }}"><svg class="svg-icon grey"><use xlink:href="{{ '/assets/minima-social-icons.svg#microdotblog' | relative_url }}"></use></svg></a></li>{%- endif -%}
   {%- if social.devto -%}<li><a href="https://dev.to/{{ social.devto | cgi_escape | escape }}" target="_blank" title="{{ social.devto | escape }}"><svg class="svg-icon grey"><use xlink:href="{{ '/assets/minima-social-icons.svg#devto' | relative_url }}"></use></svg></a></li>{%- endif -%}
+  {%- if social.google_scholar_id -%}<li><a href="https://scholar.google.com/citations?user={{ social.google_scholar_id | cgi_escape | escape }}" target="_blank" title="{{ social.google_scholar_name | escape }}"><svg class="svg-icon grey"><use xlink:href="{{ '/assets/minima-social-icons.svg#google_scholar' | relative_url }}"></use></svg></a></li>{%- endif -%}
```
juicedatom, line 21: is google_scholar_name really needed? I was mostly following the youtube_channel style convention.

juicedatom:
I added a couple questions on the code to see if anyone knows, but other than that it should be good to go!

_sass/minima/_base.scss
```diff
@@ -237,6 +237,10 @@ pre {
   color: #828282;
 }
 
+.grayscale {
```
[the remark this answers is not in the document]
juicedatom: I'm not a CSS expert though. Could be just that the svg i found doesn't like the constant `.grey` filter above. Who knows.

juicedatom:
I also tested uncommenting all the icons just to make sure things still worked.
2022-09-10-063709_1198x1856_scrot

mattr- (approved):
I'm ok with this.

@jekyll/plugin-core Any other thoughts before I merge this?

ashmaroli:
@mattr- I am on the fence regarding the use of the `grayscale()` CSS filter as against the use of `color` or `fill-color` used elsewhere.
Other than this, I am okay with the proposal.

ashmaroli:
Side note: On the whole, I really feel we should rethink about bundling social icons with the theme. Perhaps, there should be a separate project / library to allow use of social icons in Jekyll Themes like how `gemoji` provides Emoji for consumption.

juicedatom:
@ashmaroli I think that makes sense, regarding the theme.

For grey vs. grayscale I tried doing `<svg class="svg-icon grey">` but the color seemed to seep through. Is there something about the svg i'm using that doesn't support the `.grey` field?

2022-09-10-063709_1198x1856_scrot

juicedatom:
Here's an example of what happens when i just use the `grey` tag.

color

juicedatom:
Another option i guess is to completely switch over to the grayscale filter if that still works for all the different icons.

ashmaroli:
@juicedatom If you look at your SVG code, you would see multiple `fill` attributes assigned with various hexadecimal color values. I think removing all such attributes should induce adoption of the higher level `.grey` CSS rule. I am not in a position to test this myself. Hence it's just a loud thought.

juicedatom:
@ashmaroli i'll try that now.

juicedatom:
Lol I guess it technically worked.
fill

juicedatom:
Ok, I think i can mess with the opacity a bit to make it look a little nicer and keep the .grey attribute.

ashmaroli:
I took a look on the web to see if there's a good icon out there that is compatible with single color.. Unfortunately, I couldn't find one and also that even Google doesn't have easy-to-find official guidelines for this icon..

juicedatom:
Yea @ashmaroli this seems like a special case. So when i try to mess with the opacity internally I get something kind of close.
fixed-ish

to the original but based on how the paths are built opacity won't cut it. I could try messing around in the svg a bunch, but not sure it's worth it.

juicedatom:
Another alternaive is to use an out-dated icon like 
image

juicedatom:
I think i can make something work. Give me a few.

juicedatom:
Ok @ashmaroli it should be fixed now. It is using the default fill attribute.

fixed-again

ashmaroli:
Good job tweaking the code.
Could I see how it works with the dark skin of the theme?

juicedatom:
Looks a little weird in dark theme but overall not bad.

fixed-again-color-dark-theme

ashmaroli:
I think you could tweak the opacity of the circle a wee bit so that the `fill` color propagates slightly more in the dark skin without it losing integrity in the default skin. The gradients need not match pixel-by-pixel with the original icon you started with especially since Google doesn't have a strict guideline for this icon.
I'm afraid a single color will not work for this icon.

juicedatom:
Little different than the original logo but gets the job done I think.

v4-dark
v4-light

There's also a single-color version i could use, however it's an outdated logo.

v3-dark
v3-light

juicedatom:
v4-dark-grey
v4-light-grey_000
Here's the gray version of the icon (which is the default)

ashmaroli:
Let's not use the old icon for now. The current opacity looks fine. Could I see one final screenshot with default fills on the two skins?
Thanks for the extra effort put into this.

juicedatom:
Here's you'll see all the combinations of {light mode, dark mode} x {`currentColor`, `.grey` [default]}

final-default-dark
final-default-light
final-grey-dark
final-grey-light

juicedatom:
Sorry for all the editing. You'll see the default fills on the two skins in the post above. Is there anything else that I should do?

juicedatom:
Another side note: I wish that github showed the inside of svgs since they're just text files.

ashmaroli:
Looks all good to me. Thank you very much. Appreciate the effort you have put in the short duration.
GitHub does show regular standalone SVG properly. The thing is, our social asset is a composite resource of multiple SVG embedded within and that confuses the widget. I am able to see the code properly on the GitHub mobile app.

ashmaroli (commented):
LGTM! :+1:

juicedatom:
@ashmaroli oh interesting! Then is the final step for you to merge is there a bot that does it automatically?

ashmaroli:
Thanks @juicedatom
@jekyllbot: merge +minor

PR #686, opened 2022-11-11
Generate social iconsheet during build

ashmaroli:
## Summary

Simplify management and maintainability of social media icons by using multiple *include files* and individual SVG files.

## Benefits / Advantages

- Users get the ability to change the order of rendered links.
- Users get the option to render multiple links to profiles on same platform.
- Users get the option to use a different icon for a given platform.
- Users get the option to easily alter link-rendering markup.
- Theme maintainers have lower overhead over deciding to bundle each and every requested social icon.

## Drawbacks / Disadvantages

- Users have to explicitly declare a more verbose configuration to render the list-item.
- Users *may* need to edit non-builtin icons to render correctly.
- The theme no longer *escapes* user-input. The onus is entirely on the user.


mattr-:
This completely breaks backwards compatibility for folks who are using minima as a remote theme, doesn't it?

ashmaroli:
> This completely breaks backwards compatibility for folks who are using minima as a remote theme, doesn't it?

@mattr- Yes, it does. But given the fact that the theme is currently under transition state (from 2.x to 3.0), we are allowed to make breaking changes such as this. 

Currently as I see it, the benefits of this change outweighs the drawbacks. I shall update the opening comment to note the pros and cons.
Additionally, consumers of this theme as a remote-theme have the option to lock to a commit prior to a significant breaking change if not ready to move in sync with an unstable branch.

mattr- (approved):

ashmaroli:
@jekyllbot: merge +major

PR #689, opened 2022-11-17
Update google analytics README regarding G- and UA- tags

domsleee:
G'day

I think we should update the README to reflect a few things:
1. Google Analytics 4 tags (G- tags) are preferred over Universal Analytics (UA-) tags, and the [google-analytics.html snippet](https://github.com/jekyll/minima/blob/master/_includes/google-analytics.html) used in this theme works for both types
2. The Universal Analytics (UA-) tags will [stop being processed in July 2023](https://support.google.com/analytics/answer/11583528?hl=en#:~:text=Until%20July%201%2C%202023%2C%20you,for%20at%20least%20six%20months)

This information would save the next developer a few minutes of finding out if the theme supports G- tags or not, so I think it's worthwhile.


ashmaroli (approved):

ashmaroli:
Thank you @domsleee! Sorry for the delay.
@jekyllbot: merge +doc

PR #690, opened 2022-11-18
Rename `default` layout to `base`

ashmaroli:
## Rationale

- The name `default` is misleading.
- `default` generally refers to a fallback that is invoked in the absence of requested resource. However, in this case, `default` does not have a special meaning as Jekyll does not treat the layout as a special fallback layout.


ashmaroli:
Can I have your opinion on this one, @parkr?
I wonder if `layout: default` had any particular significance during the Jekyll v1-v2 era..

parkr:
Have you gotten bugs / support queries about confusion w.r.t. this name? It's been a strong convention to use this in all my Jekyll sites and has some historical value as a convention. It's not significant but the idea is to indicate that it's the "base" layout. The main drawback to this change is that it's a breaking change for all existing sites that depend on default...

parkr:
You're right that it's not a fallback. It's meant to be a base layout that all others build upon. 

We can certainly release this as a major version change but it makes it harder for systems like GitHub Pages that prize backwards compatibility above all else to upgrade :/

ashmaroli:
> Have you gotten bugs / support queries about confusion w.r.t. this name?

I did get a couple of queries from very new users wondering why their posts **not having an explicit** `layout: <some-name>` declaration in their front matter render unstyled in spite of having a proper `_layouts/default.html`. So, I thought Minima 3 would be a nice opportunity to break the convention.


> makes it harder for systems like GitHub Pages that prize backwards compatibility above all else..

I doubt GitHub Pages is going to lock-in to Minima 3 irrespective of this change. The user is free to consume this via `jekyll-remote-theme` at the user's discretion. If there is a significant push-back from the community once this merges, we can appease them with a `default` layout that inherits from the `base` layout..

parkr (approved):

mattr-:
`layout: default` has significance in that the name `default` aligns with what Rails uses as the default layout for pages in a Rails app. Happy to have it changed to something that works better for us rather than doing something just because Rails is also doing it.

ashmaroli:
@jekyllbot: merge +major

k4u5hik:
`Build Warning: Layout 'default' requested in 404.html does not exist.`

Was getting a build warning at build time. Page layout has been changed to `base` from `default`. Issue was resolved.

ashmaroli:
@k4u5hik Yes, that is a side-effect of this change. 404.html scaffolded by running `jekyll new ...` and any other "page" as such are beyond our control and requires manual intervention from users consuming the theme in current state.

PR #691, opened 2022-11-20
Footer: support RSS feed path (jekyll-feed plugin)

troy:
(This is my first commit and PR on this repo. Please feel free to reject or edit it as you see fit. Thanks!)

The jekyll-feed plugin supports an optional user-defined [RSS feed path](https://github.com/jekyll/jekyll-feed#already-have-a-feed-path). `jekyll-feed` uses that path to generate the RSS autodiscovery meta tag. This changes minima to honor that feed path when it is defined.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -6,7 +6,7 @@
     <div class="footer-col-wrapper">
       <div class="footer-col">
         <p class="feed-subscribe">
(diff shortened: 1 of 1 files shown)


ashmaroli:
Thank you @troy, but this has been superseded by #698.

PR #698, opened 2022-12-19
Improve "Subscribe" link in footer

### LINKED ISSUE #692: RSS icon depends on "rss" social_link, but "Subscribe" link does not
The [footer](https://github.com/jekyll/minima/blob/master/_includes/footer.html#L11) generates a "Subscribe" link and an RSS icon. That icon comes from a [dynamically generated](https://github.com/jekyll/minima/blob/master/assets/minima-social-icons.html) set with one icon symbol per `social_links` config entry. As far as I can tell, if `social_links` doesn't contain an `rss` entry, the RSS icon won't be included in the file.

If this explanation is accurate and there's no quick fix, perhaps add a comment/explanation in the default config file `social_links` section.

### LINKED ISSUE #697: feed.xml name hardcoded
The file name `feed.xml` is hardcoded in the "Subscribe" link. This file name is actually configurable in the jekyll-feed plugin (`feed.path`). This theme should make use of the configured value rather than hardcoding it.

ashmaroli:
## Summary

- **Respect `feed.path` setting in config file.**
     - Render feed url via `absolute_url` to mirror `{% feed_tag %}` output.
- **Always inject `rss` icon symbol in spritesheet** irrespective of config file entry.
     - Extract `symbol` template into an include file.
     - Rename spritesheet template to have `.liquid` extension for improved readability in supported text-editors / IDE.
- Remove `rel="me"` for feed link in footer.

## Context

Fixes #692
Fixes #697


mattr- (approved):

ashmaroli:
@jekyllbot: merge +fix

PR #702, opened 2022-12-29
Add a complete comment system for minima

YURLAK:
**Hello, developers!👋This is my first PR.**

## Problem.

Since I started using **minima**.I found that **minima** only comes with **Disqus** comment system.And doesn't suit for other comment system such as giscus,utterances and so on.So I have an idea that **add a complete comment system for the minima theme**.

## How to realize it?

I changed three places.

- Added  `/includes/comments.html`,Delete `/includes/disqus_comments.html`.
- Added comment settings in `_config.yml`.
- Changed **comment part** in `_layouts/post.html`.

## At last

And that's all.Thanks😀


ashmaroli:
Hello @YURLAK,
First of all, welcome to the world of Open Source Softwares.

Regarding your pull request, I love the idea but would like to leave some feedback nevertheless:
- I would prefer not to reserve a top-level config key for this to avoid conflicts on existing sites. Therefore, I'd prefer `{{ site.minima.comments.[NEW_KEY] }}` instead of the current `{{ site.[NEW_KEY] }}`.
- We no longer build the demo site using the `master` branch. So, adding the config details to `_config.yml` is not only redundant but additional information to delete for users *forking* the repository. I suggest you move all the comment configuration to the README document.
- For readability reasons, it's better that you wrap the attributes of the HTML script tags.
- Lastly, I think you forgot to include the original Disqus code into the new include file.

YURLAK:
> Hello @YURLAK, First of all, welcome to the world of Open Source Softwares.
> 
> Regarding your pull request, I love the idea but would like to leave some feedback nevertheless:
> 
> * I would prefer not to reserve a top-level config key for this to avoid conflicts on existing sites. Therefore, I'd prefer `{{ site.minima.comments.[NEW_KEY] }}` instead of the current `{{ site.[NEW_KEY] }}`.
> * We no longer build the demo site using the `master` branch. So, adding the config details to `_config.yml` is not only redundant but additional information to delete for users _forking_ the repository. I suggest you comment out all the **comment configuration** or move it to the README document.
> * For readability reasons, it's better that you wrap the attributes of the HTML script tags.
> * Lastly, I think you forgot to include the original Disqus code into the new include file.

YURLAK:
Hello,@ashmaroli.Thanks for your feedback!According to your feedback,I made some changes.

README.md
```diff
@@ -222,24 +222,45 @@ You can *add* custom metadata to the `<head />` of your layouts by creating a fi
 2. [Customize](#customization) default `_includes/custom-head.html` in your source directory and insert the given code snippet.
 
 
-### Enabling comments (via Disqus)
+### Enabling comments (via Disqus,Utterances,Giscus)
```
louisroyer, line 225:
```suggestion
### Enabling comments (via Disqus, Utterances, Giscus)
```

README.md
```diff
@@ -222,24 +222,45 @@ You can *add* custom metadata to the `<head />` of your layouts by creating a fi
...
 
-Optionally, if you have a Disqus account, you can tell Jekyll to use it to show a comments section below each post.
+If you want to add comment system for your site,add the following code to the `README.md` file.If not,skip this part.
```
louisroyer, line 227:
```suggestion
If you want to add comment system for your site, add the following code to the `README.md` file. If not, skip this part.
```

README.md
```diff
@@ -222,24 +222,45 @@ You can *add* custom metadata to the `<head />` of your layouts by creating a fi
...
+# You must install utterances github app before use.(https://github.com/apps/utterances)
+# Make sure all variables are set properly. Check below link for detail.
+# https://utteranc.es/
+utterances:
+  repo:                "username/username.github.io"
+  issue-term:          "pathname"
+  label:               "Comments"
+  theme:               "your_theme"
+
+# You must install giscus github app before use.(https://github.com/apps/giscus)
```
louisroyer, line 248:
```suggestion
# You must install giscus github app before use (https://github.com/apps/giscus).
```

README.md
```diff
@@ -222,24 +222,45 @@ You can *add* custom metadata to the `<head />` of your layouts by creating a fi
...
+  category:            "your_category"
+  category-id:         "your_category_id"
+  mapping:             "pathname"
+  reaction-enabled:    "1"
+  theme:               "your_theme"
+  crossorigin:         "your_crossorigin"
+  lang:                "your_language"
+ ```
+ 
+ :warning: First,you need to enter the name of your comment system into the `provider` option(`giscus`,`utterances` or `disqus`).Second,enter the configurations of your comment system.
```
louisroyer, line 263:
```suggestion
 :warning: First, you need to enter the name of your comment system into the `provider` option (`giscus`, `utterances` or `disqus`). Second, enter the configurations of your comment system.
```

louisroyer (changes requested):
Some typographic suggestions.

ashmaroli:
@YURLAK I thought about this bit more. Continuing from previous feedback:
- Comments are not handled by Jekyll itself so no point in having it as a top-level config key.
**Let's use `{{ site.minima.comments }}`** instead.
- Rendering comments during development is not desired as it creates unnecessary discussions on Disqus. Perhaps, the same is true for Giscus and Utterances. Either ways, render comments only in `"production"`.
- The script rendered is easily accessible on provider's website. Therefore, we can avoid performance issues on larger sites by recommending users to copy from provider site (or our README) and *hardcode* user details as needed. We just provide a means to render desired provider script.
- Additionally, provide a means to disable comments entirely for desired posts (via `page.comments`).
- No built-in comment support to reduce coupling to third-party code. 

So, our `_layouts/post.html` will have something like:

```
{% if jekyll.environment == "production" %}
  {% unless page.comments == "disabled" %}
    {% assign comments_provider == site.minima.comments_provider %}
    {% if comments_provider %}
      {% capture provider_filepath %}comments/{{ comments_provider }}.html{% endcapture %}
      {% include {{ provider_filepath }} %}
    {% endif %}
  {% endunless %}
{% endif %}
```
Then, we document what users need to do, in our README along with the note that build will fail if user does not provide associated include file.

PR #713, opened 2023-02-20
Fix reference to base layout in `404.html`

### LINKED ISSUE #712: `404.html` still references layout `default.html`
layout `default.html` was renamed to `base.html`, but `404.html` still references the old name.

ashmaroli:
Update `404.html` to use `base.html` layout instead of `default.html` layout.

Fixes #712

### DIFF
--- a/404.html
+++ b/404.html
@@ -1,5 +1,5 @@
 ---
-layout: default
+layout: base
 ---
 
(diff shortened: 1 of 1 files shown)


ashmaroli:
@jekyllbot: merge +fix

PR #718, opened 2023-03-09
Add noheader

Janorico:
Add noheader front matter variable. If true, the page was not be displayed at the site header.


ashmaroli:
If you check the `_config.yml` at the root of this repository, you'll see a commented-out `header_links` key. We use that key to support custom header entries.

PR #719, opened 2023-03-10
Highlight that the base branch is v3

ashmaroli:
Many visitors do not realize that the README on `master` documents unreleased features pertaining to next major version, v3.


ashmaroli:
@jekyllbot: merge +doc

PR #760, opened 2023-11-13
Improve documentation for config options

Because789:
This update to _config.yml tries to avoid some confusion I ran into after choosing Minima as the theme for my GH Pages site. I did not want to build locally, but let GitHub do the job. Here are the changes I made:

- Since GH Pages still uses Minima 2.5.1, I added a block on how to use the latest Minima version anyway. 
- My header_pages did not show up in the specified order. It was an indentation error, for some reason I thought that it is a minima setting (minima.header_pages). I placed it before the minima block to avoid this confusion.
- The skin setting is missing in the current default _config.yml, I added it here.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -12,11 +12,11 @@ description: >
 
 # Build settings
 
-# If you clone the minima repo and build locally, use this setting.
+# If you clone the Minima repo and build locally, use this setting.
 theme: minima
 
-# As of November 2023, GitHub Pages still uses minima 2.5.1 (https://pages.github.com/versions/). 
-# If you want to use the latest minima version on GitHub Pages, use the following setting and 
+# As of November 2023, GitHub Pages still uses Minima 2.5.1 (https://pages.github.com/versions/). 
+# If you want to use the latest Minima version on GitHub Pages, use the following setting and 
 # add a line comment on "theme: minima" above.
 #remote_theme: jekyll/minima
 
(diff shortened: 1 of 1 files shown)


_config.yml
```diff
@@ -12,46 +12,61 @@ description: >
...
+  # solarized-light	   Light variant of solarized color scheme.
+  # solarized-dark	   Dark variant of solarized color scheme.
+  # solarized	         Adaptive skin for solarized color scheme skins.
+  skin: classic
+  
   # Minima date format.
   # Refer to https://shopify.github.io/liquid/filters/date/ if you want to customize this.
-  #
-  # date_format: "%b %-d, %Y"
+  date_format: "%b %-d, %Y"
```
mattr-, line 51: I didn't see this change described in your PR description. Was it intentional to uncomment this? If not, would you mind recommenting it? Thanks!

mattr- (commented):
I'm on board with this in general. I left one inline comment.

Would you mind updating the PR title to be a short sentence describing your changes? PR titles get used in `History.markdown`. Something like `Improve documentation for config options` or something like that would be ideal.

Thanks for your contribution! :heart:

_config.yml
Because789: Yes, I forgot about that, sorry, it was late. The idea was, that the config file should reflect the sample site (https://jekyll.github.io/minima/). I just tested it, and it seems that the default date value is "%b %d, %Y" anyway (I couldn't find out what the minus before the day means, it seems to do nothing). I would therefor suggest, to comment the line again, but switch it to a format which is not the default (we could mention the default in the comment above).

mattr- (approved):
Thanks for your work here! ❤️ 

@jekyllbot: merge +docs

PR #764, opened 2023-12-01
add `mastodon` to social-media platforms in README

goseind:
I noticed the `.svg` was added already but Mastodon not listed under the social-media platforms in the README. This is a minor thing but might help new users to know it's there 👀


mattr-:
Thanks for contributing! This was superceded by #788

PR #766, opened 2023-12-08
Add email icon

peterbozso:


_config.yml
```diff
@@ -55,6 +55,7 @@ minima:
   #social_links:
   #  - { platform: devto,          user_url: "https://dev.to/jekyll" }
   #  - { platform: dribbble,       user_url: "https://dribbble.com/jekyll" }
+  #  - { platform: email,          user_url: "mailto:minima@jekyll.com" }
```
mattr-, line 58: Since this isn't a real email address, I'd prefer to go with a slightly different approach here
```suggestion
  #  - { platform: email,          user_url: "mailto:jekyll_minima@example.com" }
```

ashmaroli:
Thank you for the pull request but social icons are now rendered via the Font Awesome library.
Email icon doesn't belong in the list of social platforms though.

PR #775, opened 2024-01-31
Minimal fixes for Disqus

kant2002:
I create local modification, and only using absolute URL make Disqus works.

### DIFF
--- a/_includes/disqus_comments.html
+++ b/_includes/disqus_comments.html
@@ -3,7 +3,7 @@
   <div id="disqus_thread"></div>
   <script>
     var disqus_config = function () {
-      this.page.url = '{{ page.url | absolute_url }}';
+      this.page.url = '{{ absolute_url }}';
       this.page.identifier = '{{ page.url | absolute_url }}';
     };
 


kant2002:
We have #536 and #569 which attempt to fix default. I think that's all what is needed. Disqus only wants absolute URL

mattr- (approved):
Thank you for the contribution!

@jekyllbot: merge +fix

_includes/disqus_comments.html
```diff
@@ -3,7 +3,7 @@
   <div id="disqus_thread"></div>
   <script>
     var disqus_config = function () {
-      this.page.url = '{{ page.url | absolute_url }}';
+      this.page.url = '{{ absolute_url }}';
```
revolter, line 6: `absolute_url` is a filter (https://jekyllrb.com/docs/liquid/filters/), and, from my knowledge and testing, it doesn't do anything when used directly.

So, shouldn't this be changed to
```js
this.page.url = '';
```
?

kant2002: I think this did not work for me on GitHub pages.

revolter: What does `{{ absolute_url }}` expand to, in your case?

I tested, and in my case, it expands to nothing: `this.page.url = '';`, which is what I assumed.

kant2002: You are right. in my case it's `this.page.url = '';`

Me042589:
Woke

PR #677, opened 2022-09-18
Fix responsiveness of social SVG sprites

ashmaroli:
### Summary

- SVG icons require  `viewBox` attribute to be responsive.
- Remove unnecessary CSS to normalize appearance.

### Additional Changes

This PR also slightly increases the *overall render area* of the icons for better appearance on mobile devices.

PR #808, opened 2024-09-26
Avoid clipping responsive social SVG icons

caizixian:
After #677 , social icons slightly clips, probably due to issues similar to https://gsap.com/community/forums/topic/16844-svg-overflowing-its-bounding-box-while-scaling/

For example, see the right edge of the GitHub icon 
image

The CSS property added in this PR fixes it.
image

### DIFF
--- a/_sass/minima/_base.scss
+++ b/_sass/minima/_base.scss
@@ -243,6 +243,7 @@ pre {
   display: inline-block;
   fill: currentColor;
   vertical-align: text-bottom;
+  overflow: visible;
 }
 
 


ashmaroli (approved):
While I still don't encounter the clipping at my browser's default setting, I do now see the clipping but at a different magnification.
So, any argument to justify my lapse in #677 is moot.

This is a valid fix.

ashmaroli:
Thank you @caizixian

P.S. @caizixian Do you also encounter a *flaky* situation where sometimes the social-icons do not render at all in spite of being configured properly in the config file and include_file?

@jekyllbot: merge +fix

caizixian:
> Thank you @caizixian
> 
> P.S. @caizixian Do you also encounter a _flaky_ situation where sometimes the social-icons do not render at all in spite of being configured properly in the config file and include_file?
> 
> @jekyllbot: merge +fix

Sorry that I somehow missed this. I haven't encountered this issue.

PR #812, opened 2024-09-30
corrections for deprecation warning: "Deprecation Warning: lighten() …

brian-davis:
…is deprecated.", use Sass color module


ashmaroli:
Thank you for the proposal, @brian-davis but unfortunately, this change is not compatible with Jekyll 3.10 (GitHub Pages users via remote_theme). So, this cannot be merged right away. Will you be able to come up with a solution that is compatible with both old Ruby Sass API and modern Dart Sass API?

PR #840, opened 2025-02-09
Use single-quotes when nested within double-quoted attributes

ashmaroli:
* Easier to read.
* Avoids being highlighted as an error in some text-editors.


ashmaroli:
@jekyllbot: merge +fix

PR #841, opened 2025-02-09
Move `show_excerpts` setting to be under `minima` namespace

ashmaroli:
`show_excerpts` is not a configuration key defined by Jekyll. So move it to be under the `minima` namespace.


ashmaroli:
@jekyllbot: merge +major

Issue #843, opened 2025-02-12
Social platform icons now restricted to Font Awesome library fa-brands


pfirpfel:
Regarding #839:

I understand that in most cases, this feature is now much simpler to use. But you are limited on what **brand** icons fontawesome has.

Would it be possible to still support the old icon system as a fallback?
Or not limit the icons to the fa-brands namespace in fontawesome? In my case being able to switch to `fa-regular` instead of `fa-brands` would be enough.

ashmaroli:
We don't intend to support the old icon system alongside the current Font Awesome based system due to reasons mentioned in #839. Limiting support to just `fa-brands` was a conscious decision.

However, I am curious as to which `fa-regular` icons you intend to render as part of the *social-icons* list. If its the `envelop` icon to denote `email`, then I believe that is duplicating the info rendered elsewhere in the footer itself. You are free to customize the `_includes/footer.html` to change the rendering of footer or just `_includes/social.html` at your end if you want to render the `envelop` icon as part of *social-links*.

pfirpfel:
I used to have more or less `fa-calendar` as an icon. But then it would need `fa-brands` changed into `fa-regular`. Same for `fa-envelope`.

I guess right now, the only way is to override the includes as you said. Being able to override the fontawesome "namespace" (`fa-brands` being the default) per social item would be cool though.

pfirpfel:
This is how I solved it in my `social.html` for now:

```html
<span class="grey fa-{{ entry.fa_namespace | default: "brands" }} fa-{{ entry.icon }} fa-lg"></span>
```

Config:
```yaml
minima:
  social_links:
    - title: Jekyll repository at GitHub
      icon: github
      url: "https://github.com/jekyll/jekyll"
    - title: Jekyll at X (formerly Twitter)
      icon: x-twitter
      url: "https://x.com/jekyllrb"
    - title: Email Jekyll
      icon: envelope
      fa_namespace: regular
      url: "mailto:jekyll@example.com"
````

harisont:
> However, I am curious as to which fa-regular icons you intend to render as part of the social-icons list. If its the envelop icon to denote email, then I believe that is duplicating the info rendered elsewhere in the footer itself. You are free to customize the _includes/footer.html to change the rendering of footer or just _includes/social.html at your end if you want to render the envelop icon as part of social-links.

[...]

PR #839, opened 2025-02-06
Render social platform icons via Font Awesome library

ashmaroli:
## Summary

Load **latest Font Awesome Free webfonts** via **remote CDN**.

## Interface changes

(_w.r.t. current `master`_)

- Config keys under `minima:social_links` are now:
  - `icon` - Font Awesome icon id. e.g. `icon: github` corresponds to FA icon `fa-github`.
  - `url` - Full social profile URL.
  - `title` - link-title rendered onMouseOver icon.
- Manage via a single include_file: `_includes/social.html`. The following are **removed**:
  - `_includes/social-icons/*.svg` 🔥
  - `_includes/social-item.html` 🔥
  - `_includes/svg_symbol.html` 🔥
  - `assets/minima-social-icons.liquid` 🔥

*N.B: icon for RSS is now inlined into HTML markup.*

## Motivation

- Reduce end-user development complexity associated with existing flow (on `master`).
- Remove Maintainer overhead on managing pull requests and issue tickets regarding social-icons.

PR #844, opened 2025-02-13
Add optional fa_icon_pack property to social_links to allow more Font Awesome icons packs beyond brands

pfirpfel:
# Summary
Extends functionality introduced in #839. Proposal for my issue raised here: #843

# Interface changes
Adds an optional property to the config keys under minima:social_links. Will not break existing configurations.

# Motivation
Allow users to use icons from Font Awesome beyond the brands icon pack in their social links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,10 +294,16 @@ minima:
     - title: Jekyll at X (formerly Twitter)
       icon: x-twitter
       url: "https://x.com/jekyllrb"
+    - title: Email Jekyll
+      icon: envelope
+      fa_icon_pack: regular
+      url: "mailto:jekyll@example.com"
 ```
-
-where `title` corresponds to the link-title displayed when a visitor hovers mouse-pointer over url / icon and
-`icon` refers to the Font Awesome icon id. e.g. `github` corresponds to `fa-github`.
+Properties:
+- `title` corresponds to the link-title displayed when a visitor hovers mouse-pointer over url / icon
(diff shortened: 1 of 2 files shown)


ashmaroli:
Hello again, @pfirpfel
I am not entirely in favor of using Liquid's `default` filter in a for loop owing to potential performance issues with using the theme in a site with numerous posts and pages because of [this conditional flow](https://github.com/Shopify/liquid/blob/ae3057e94b7c4d657e6bc02e1d50398e34cc6ed7/lib/liquid/standardfilters.rb#L423). <br/>(*imagine calling `:respond_to?` for every social-icon entry with `fa_icon_pack` property set to a non-falsey value for every post / page in the site.. will it be too much of a concern? Maybe, maybe not.*)

The one viable solution I see here is to force *everyone* to set the *entire* necessary Font Awesome icon classlist:
```yaml
minima:
  social_links:
    - title: Jekyll repository at GitHub
      icon: "fab fa-github"
      url: "https://github.com/jekyll/jekyll"
    - title: Email Jekyll
      icon: "far fa-envelope"
      url: "mailto:jekyll@example.com"
```
Note how one can now opt to use `fab` or `far` instead of the expanded forms `fa-brands` and `fa-regular` respectively if they are familiar with that syntax.

---

N.B. You need not rebase your branch if you choose to make changes. We use squash merges here.

ashmaroli:
Any thoughts on my suggestion above, @pfirpfel?

PR #846, opened 2025-02-22
Silence Jekyll build warning concerning logger dependency

chaoscommencer:
WARNING: "logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0. You can add logger to your Gemfile or gemspec to silence this warning.


ashmaroli:
Hello @chaoscommencer, thank you for submitting this pull request. However, this is outside the scope of a jekyll-theme.
This will be automatically resolved when the next version of `mercenary` gem will be released.

PR #847, opened 2025-02-22
Address Sass '@import' deprecation warnings

chaoscommencer:
Deprecation warnings:
- Deprecation Warning [import]: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.
  More info and automated migrator: https://sass-lang.com/d/import

Migrate to use of '@use', '@forward', and '@include' instead

NOTE: Required migration of common code to a distinct file that could be shared (via @use) by the other sass files


chaoscommencer:
I apologize; these changes were supposed to be applied to 2.5-stable

PR #850, opened 2025-02-22
Integrate common icon base

chaoscommencer:
- Isolate a common icon base include for representing icons
- Extract individual icon code to distinct files
- Integrate common icon base into each icon file

NOTE: Additionally removes unused SVG icons for GitHub and Twitter


ashmaroli:
Hello @chaoscommencer, thank you for submitting this pull request. However, this is a *non-backwards compatible* change and therefore not within the scope of the 2.x series.

A similar approach was implemented in the `master` branch but has now dropped that in favor of rendering icons from Font Awesome Free library.

chaoscommencer:
Ok, thanks for the update!

PR #852, opened 2025-02-23
Simplify `.wrapper` style definition

ashmaroli:
- max-width of **child elements** of `.wrapper` now equals `$content-width` &mdash; no more complicated math needed to determine width of *content text*.
- consistent padding at all viewport sizes.


ashmaroli:
@jekyllbot: merge +major

PR #853, opened 2025-02-24
better description of disqus and google analytics setup

Polovinkin:
from the wording it wasn't very clear where exactly should users put the lines for configuration (Jekyll site doesn't say much), added some details.
Also Google Analytics id now looks different and starts with G instead, also changed


Polovinkin:
@mattr- hi, yeah I kinda forgot about that, will try to make those fixes on this week

PR #854, opened 2025-02-24
Simplify inserting markup before closing body tag

ashmaroli:
Introduce new inclusion / include_file named `sub-footer.html` to allow inserting markup before the closing `</body>` tag.
- No need to customize `_layouts/base.html`.
- No need to customize `_includes/footer.html`.


ashmaroli:
@jekyllbot: merge +minor

PR #857, opened 2025-03-03
Improve link underline styling

ashmaroli:
- Add an offset between glyph and underline
- Render underline with `$brand-color` instead of `$link-hover-color`.

### Before

underline-before

### After

underline-after


ashmaroli:
@jekyllbot: merge +fix

PR #865, opened 2025-03-16
Reduce wrapper padding on small screens

ashmaroli:
Make better use of limited view-space on small screens.
Was `15px !default` in v2.5.
Is currently `30px !default` on `master`.

Now it will be `18px !default`.


ashmaroli:
@jekyllbot: merge +fix

PR #866, opened 2025-03-17
Render headings and text in different colors

ashmaroli:
To be able to visually distinguish paragraph-text and headings, especially `h4, h5, h6` that have similar font-sizes to `p` elements.


ashmaroli:
@jekyllbot: merge +minor

PR #868, opened 2025-03-19
Set default `font-weight` of headings to `500`

ashmaroli:
But retain existing style for `.post-title`.


ashmaroli:
@jekyllbot: merge +major

PR #869, opened 2025-03-19
Adjust table-text color based on heading-color

ashmaroli:
To be the same color prior to change in cf7e884


ashmaroli:
@jekyllbot: merge +fix

PR #878, opened 2025-03-28
Toggle icons based on `#nav-trigger` state

ashmaroli:
Instead of having the icon for `#nav-trigger` remain *the triple-bars* / *hamburger* regardless of having the mobile navigation displayed, toggle between the triple-bars and ✖️ based on the state of the trigger.


ashmaroli:
@jekyllbot: merge +fix

PR #884, opened 2025-04-04
Undo fixed header

ashmaroli:
Revert decision to have the `.site-header` fixed to top of viewport because of the following UX issue:
  - Fixed header creates confusion when user navigates to an anchor (especially when the anchor is not at start of the page) because the anchor (typically a heading element) gets hidden underneath the sticky-header.


ashmaroli:
@jekyllbot: merge +fix

PR #885, opened 2025-04-04
Render minimal site-nav on small screens

### LINKED ISSUE #860: Navigation on mobile is now much worse visually because of full-height element
Happened after this PR was merged, and in my opinion it's the complete opposite of it's title "Improve navigation styling and experience": https://github.com/jekyll/minima/pull/856

I think that making this visual element filling all the height of the screen is completely unnecessary and makes visual experience much worse for mobile users, I can imagine most navigations have just several items there, why expand this and have a huge window for almost half a screen which for most websites will be like 80% empty? What was the problem with small window approach which we had before?

ashmaroli:
Resolves #860

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -8,19 +8,12 @@
 
     {%- unless page_titles == empty %}
       <nav class="site-nav">
-        <input type="checkbox" id="nav-trigger" class="nav-trigger" />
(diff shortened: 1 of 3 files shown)


ashmaroli:
@jekyllbot: merge +fix

PR #886, opened 2025-04-07
Render comments via platform-agnostic inclusion

ashmaroli:
- Move markup to render *disqus comments* into `_includes/comments.html`. Rather than renaming the include_file, the following changes have also been made:
  - move `JEKYLL_ENV=production` and `comments: false` front matter checks up into `post` layout.
  - move check for `site.disqus.shortname` into `_includes/comments.html`.
  - `this.page.identifier` javascript variable is set to `{{ page.id }}`.
- Render a **`Comments have been disabled for this post.`** message in posts with `comments: false` in front matter.

### Rationale

A generic inclusion / include_file name would mean a user could opt to another comments-provider-platform by editing the generic include_file rather than editing the `post` layout as well as introducing a new include_file at user's source-dir.


ashmaroli:
@jekyllbot: merge +major

IvanUkhov:
I think this PR broke the comments due to this line removed:

```html
```

The embedded script is looking for that ID and fails to find. Is it supposed to be added manually after this change, or was it overlooked?

PR #891, opened 2025-04-21
Adjust colors based on WCAG contrast criterion

ashmaroli:
`Web Content Accessibility Guidelines 2.0 level AA` recommends a minimum contrast-ratio of `4.5:1` between foreground-color and background-color for normal-sized text.

Hence adjust colors of various skins in accordance with this recommendation.


ashmaroli:
@jekyllbot: merge +minor

PR #893, opened 2025-04-25
Improve rendering of pagination pagers

ashmaroli:
- Fixes breaks in alignment of pagers that crept in since the time pagers were originally styled.
- Improves accessibility by adding `title` attributes for link elements in pagers.
- `.current-page` element now sports a bold font-weight.


ashmaroli:
@jekyllbot: merge +minor

PR #894, opened 2025-04-25
Remove double title from md

cychitivav:


ashmaroli:
Hello @cychitivav, please provide a description regarding this change in the opening post. I am not able to understand the rationale behind this proposal.

cychitivav:
When using this template with GitHub Pages, the page title is rendered twice if the markdown file includes a level-one heading (`# My Title`).

This happens because:
1. The template automatically renders the title using `{{ page.title }}`.
2. The content still includes the same heading, resulting in a duplicated title.

ashmaroli:
The root cause behind the double-render is the plugin `jekyll-titles-from-headings` that is automatically included and activated in GitHub Pages build environment.

I suggest that you try the [`strip_title` configuration](https://github.com/benbalter/jekyll-titles-from-headings?tab=readme-ov-file#stripping-titles) documented in the plugin docs instead of modifying the theme template.

cychitivav:
Thank you

